PR #14, opened 2022-12-08
WIP: Add content health dashboard

pipermerriam:
WIP


pipermerriam:
- get static file serving going for devopment environments: https://github.com/tokio-rs/axum/blob/main/examples/static-file-server/src/main.rs
- get some basic styling in place using css framework. (which one?)
- sketch out a basic health overview dashboard, list view for both content-id and content-key, and then detail views for content audits.

PR #15, opened 2022-12-08
Glados Audit Process

mrferris:
- [x] Periodically select content keys
- [x] Send working content query
- [x] Create `contentaudit` entry in DB for the query response
- [x] Add timestamp to ContentKey and get most recent n content keys


mrferris:
The basic idea is working: every 2 minutes, the 10 most recent blocks are audited sequentially.

Currently working on modifying `PortalClient` to allow for parallel content queries.

pipermerriam:
Suggestion on how to do the selection of what to audit. 

I believe we should be able to construct a query that orders things by the following two criteria:

- number of times something has been audited (zero audits first, multiple audits last)
- date the content was "minted" (I'm using minted instead of created to signify that we care about the moment the data was first available on chain, not when we created it in our database). (most recent first, oldest last)

I think that this correctly prioritizes things such that it will consistently focus on the most recent data, but will work its way backwards into older data as well.  No need to do this on the first iteration, but I think it's the direction we want to go towards.

mrferris:
Deferring batched, concurrent audits until a further PR, after trin's JSONRPC has been modified to execute queries concurrently.

PR #18, opened 2022-12-13
glados-monitor: `continue` checking for new blocks instead of `break`ing out of loop

mrferris:
`glados-monitor` stopped finding new blocks after 3-4 blocks because it was hitting this `break`


pipermerriam:
I think you can simply strip out the whole `if` block there.

PR #26, opened 2022-12-19
add content keys for block body and receipts

pipermerriam:
Adds creation and tracking of content keys for the block body and block receipts


pipermerriam:
There is a bit of redundant logic in the different implementations here, specifically the `Display` implementation and the `ContentKey::encode()` implementations are all actually "generic" but I couldn't get the type system to agree with me.  Going to see about pairing with someone who knows rust better on how to remove this code duplication.

PR #28, opened 2022-12-21
Add audit details page and audit tracing

mrferris:
This PR adds a string column to the `contentaudit` schema, and renders the query graph on the audit details page for audits that have traces.

Screen Shot 2023-01-04 at 1 08 54 PM

todo:

- [x] Add query stats to audit details page
- [x] Add table view of nodes next to graph view
- [x] Get rid of 13k line diff to `bootstrap.min.css`

### DIFF
--- a/entity/src/test.rs
+++ b/entity/src/test.rs
@@ -213,6 +213,7 @@ async fn test_audit_crud() -> Result<(), DbErr> {
         created_at: Set(Utc::now().into()),
         strategy_used: Set(Some(SelectionStrategy::Random)),
         result: Set(content_audit::AuditResult::Success),
+        trace: Set("{}".to_owned()),
     };
 
     let content_audit_model = content_audit_active_model.insert(&conn).await?;
--- a/entity/src/test.rs
+++ b/entity/src/test.rs
@@ -213,7 +213,7 @@ async fn test_audit_crud() -> Result<(), DbErr> {
         created_at: Set(Utc::now().into()),
         strategy_used: Set(Some(SelectionStrategy::Random)),
         result: Set(content_audit::AuditResult::Success),
-        trace: Set("{}".to_owned()),
+        trace: Set("".to_owned()),
     };
 
     let content_audit_model = content_audit_active_model.insert(&conn).await?;
--- a/glados-web/templates/contentaudit_detail.html
+++ b/glados-web/templates/contentaudit_detail.html
@@ -12,10 +12,6 @@
 <div id="top-of-audit-detail">
     <div class="row">
         <h1>Audit #{{ audit.id }}</h1>
-        <span>
-            <a href="/audit/id/{{ audit.id - 1 }}" class="previous">&laquo; Previous</a>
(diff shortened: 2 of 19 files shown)


mrferris:
@pipermerriam When you have a second mind quickly reviewing this? The DB migration is the main thing that will affect anything else. 

I've updated this to not call the `Trace` endpoint, so trace columns in the DB will be empty and things will looks the same. 

Once we merge #110 we'll know on the fly whether a given client supports the `Trace` endpoint and if so we'll populate the trace field and the functionality in this PR will link & display the query's trace.

entity/src/test.rs
```diff
@@ -213,6 +213,7 @@ async fn test_audit_crud() -> Result<(), DbErr> {
         created_at: Set(Utc::now().into()),
         strategy_used: Set(Some(SelectionStrategy::Random)),
         result: Set(content_audit::AuditResult::Success),
+        trace: Set("{}".to_owned()),
```
pipermerriam, line 216: Just curious why this is using `"{}"` and I'm seeing just empty string in other places.

glados-web/templates/contentaudit_detail.html
```diff
@@ -0,0 +1,139 @@
...
+<script src="/static/js/trace/graph.js"></script>
+<script src="/static/js/trace/main.js"></script>
+
+
+<div id="top-of-audit-detail">
+    <div class="row">
+        <h1>Audit #{{ audit.id }}</h1>
+        <span>
+            <a href="/audit/id/{{ audit.id - 1 }}" class="previous">&laquo; Previous</a>
+            <a href="/audit/id/{{ audit.id + 1 }}" class="next">Next &raquo;</a>
```
pipermerriam, line 17: I'm somewhat :-1: on this because it works off of the assumption that we have all ID values in database...  I don't think we need next/prev buttons.

glados-web/templates/contentaudit_detail.html
```diff
@@ -0,0 +1,139 @@
...
+        <div class="col">
+            <ul>
+                <li>Content Key: {{ content.key_as_hex() }}</li>
+                <li>Content ID: {{ content.id_as_hex() }}</li>
+                <li>Started: {{ audit.created_at }}</li>
+                <li>Result: {{ audit.result.as_text() }}</li>
+            </ul>
+        </div>
+    </div>
+    <div id="no-trace" hidden="true">No Trace Available</div>
```
pipermerriam, line 28: Seems odd to do this with `hidden="true"` rather than using template `{% if ... %}` blocks...  Curious if there's a reason you went this route.

pipermerriam (approved):
:+1:

glados-web/templates/contentaudit_detail.html
mrferris: Yea, I'd agree with that.

The way it's currently done is that any exception in the javascript code results in "No Trace Available" being shown, empty trace included.

When I add the call to content tracing after #110 I'll make this cleaner.

PR #29, opened 2023-01-21
Error logging

perama-v:
## Description

Panic when running `glados-monitor` against a local execution node.
```sh
$ cargo run -p glados-monitor -- --database-url sqlite::memory: follow-head --provider-url http://127.0.0.1:8545
```
The panic occurred in `contentkey::get_or_create()`, where an `Entity::find` Result was `unwrap()`ed.

### Monitor loops do not terminate
The `run_glados_monitor()` initiates two never-ending loops:
- `follow_chain_head()`
- `retrieve_new_blocks()`

Rather than allow those loops to fail when an error is encountered, the loops are allowed to continue. 

- Errors are logged
- Where the loop cannot proceed due to an error, a `continue` moves to the next cycle.

### Channel remains open

The `retrieve_new_blocks()` control loop was determined by a `while let` on the `channel.recv()`. 

After the first successful loop, the function terminated and the channel was closed. This was resolved
by replacing the `while` with a `loop`.

### Early terminations

Some conditions allow for a loop to be skipped early. Those were moved to `continue`s to reduce block
nesting.

### Grouping of `ContentKey` storing

The logic for storage of the three `ContentKey` types was moved to a function generic over that type.

## Unresolved

There is still an issue where the following function results in an `Error`
```rs
  let content_key = Entity::find()
      .filter(Column::ContentKey.eq(content_key_raw.encode()))
      .one(conn)
      .await?;
```
```sh
error: Query Error: error returned from database: (code: 1) no such table: content_key
```
I am not sure what this problem is. However rather than crashing glados, it runs its loops and logs the error.

### Additional changes
- Added example flag in docs to clarify that glados-monitor connects to

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,12 +48,15 @@ In most cases, you will want to set the environment variable `RUST_LOG` to enabl
 
 The `glados-monitor` crate can be run as follows to populate a local database with content ids.
 
-The CLI needs a DATABASE_URL to know what relational database to connect to, as well as an HTTP_PROVIDER_URI to connect to an Ethereum JSON-RPC provider.
+The CLI needs a DATABASE_URL to know what relational database to connect to, as well as an HTTP_PROVIDER_URI to connect to an Ethereum JSON-RPC provider (not a Portal node).
 
 ```
 $ cargo run -p glados-monitor -- --database-url <DATABASE_URL> follow-head --provider-url <HTTP_PROVIDER_URI>
 ```
-
+For example, if an Ethereum execution client is running on localhost port 8545:
+```
+$ cargo run -p glados-monitor -- --database-url sqlite::memory: follow-head --provider-url http://127.0.0.1:8545
+```
 #### Importing the pre-merge accumulators
 
 The pre-merge epoch accumulators can be found here: https://github.com/njgheorghita/portal-accumulators
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ In most cases, you will want to set the environment variable `RUST_LOG` to enabl
 
 The `glados-monitor` crate can be run as follows to populate a local database with content ids.
 
-The CLI needs a DATABASE_URL to know what relational database to connect to, as well as an HTTP_PROVIDER_URI to connect to an Ethereum JSON-RPC provider (not a Portal node).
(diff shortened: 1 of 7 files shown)


glados-monitor/src/lib.rs
```diff
@@ -86,77 +84,75 @@ async fn retrieve_new_blocks(
...
+        content.id=?key.content_id(),
+        content.kind=name,
+        "Creating content database record",
+    );
+
+    if let Err(e) = contentkey::get_or_create(key, conn).await {
+        error!(
+            "Failed to record {} key in db (key = {}, id = {}) (error: {})",
+            name,
+            key.hex_encode(),
```
pipermerriam, line 145: @jacobkaufmann pinging you because I think you might know this better than me.

I think our preference here is to use structured logging over string formatting.  You can see in the `debug!(..)` call a few lines earlier how we can dive each of the pieces of data a key and value and the structured logging machinery naturally puts them in as key/value pairs.....

Did I get that right Jacob?

perama-v: Ok happy to update to use structured logs. Is there a good way to know what to make the keys? 

In the debug log above `content.kind` is used, but maybe it could have been `content.name` with no ill effect? 

Is the idea that there is a system, which connects to a log stream and allows rapid filtering of a set of common keys? E.g. filter for logs containing key `content.kind` that would be broken if `content.name` had been used.

perama-v: (done)

README.md
```diff
@@ -48,12 +48,15 @@ In most cases, you will want to set the environment variable `RUST_LOG` to enabl
 
 The `glados-monitor` crate can be run as follows to populate a local database with content ids.
 
-The CLI needs a DATABASE_URL to know what relational database to connect to, as well as an HTTP_PROVIDER_URI to connect to an Ethereum JSON-RPC provider.
+The CLI needs a DATABASE_URL to know what relational database to connect to, as well as an HTTP_PROVIDER_URI to connect to an Ethereum JSON-RPC provider (not a Portal node).
```
pipermerriam, line 51:
```suggestion
The CLI needs a DATABASE_URL to know what relational database to connect to, as well as an HTTP_PROVIDER_URI to connect to an Ethereum JSON-RPC provider (not a portal node).
```

Issue #30, opened 2023-01-23
Database is not initialized for in-memory db


perama-v:
### Description

In `glados-monitor`, when an in-memory database is created using the `"sqlite::memory:"` argument, as in:

```command
$ cargo run -p glados-monitor -- --database-url sqlite::memory: follow-head --provider-url http://127.0.0.1:8545
```
The following error is encountered:
```sh
Query(SqlxError(Database(SqliteError { code: 1, message: "no such table: content_key" })))
```
path
glados/entity/src/contentkey.rs:46:10
```
This indicates that upon lookup of the first key in the table, the action fails as the result of a non-existent table. 
```rs
// glados/entity/src/contentkey.rs
get_or_create(content_key_raw: &impl ContentKey, conn: &DatabaseConnection) -> Model {
    // First try to lookup an existing entry.
    let content_key = Entity::find()
        .filter(Column::ContentKey.eq(content_key_raw.encode()))
        .one(conn)
        .await
        .unwrap(); <--- panic
    // snip
}
```
## Resolution
Some options is see:
1. Require that the CLI flags prevent in-memory-db + no-migration combination
2. Run db migration if in-memory-db is selected (this initialises the db)

mrferris:
Is this issue specific to in memory DB? I'd think this same thing happens for a sqlite DB.

I think that both options 1 & 2 would remove the control that the `migrate` flag gives us.

I propose options:

3. Detect whether there's any DB, if so always run `Migrator::up`. If a DB exists, check the `migrate` flag.
4. Just pass `--migrate` in every time, unless you specifically don't want to migrate.
5. Replace `--migrate` with `--no-migrate` and make `Migrator::up` the default behavior.

perama-v:
Good point, I've updated #31 to do `3` instead. The use of the migration flag for a new or in-memory database (`4`) feels counter-intuitive and agree on `5`.

perama-v:
On discussion in the PR, favouring `4` at this point, with the addition that in-memory db always triggers `Migrator::up()`

perama-v:
Closed by #31

PR #31, opened 2023-01-23
fix: Don't skip db init for in-memory db

perama-v:
### Description

No database is created for in-memory db as noted in #30 

### What was changed

- ~~Implements option 2 outlined in the issue by detecting if the in-memory argument was passed and a running db migration to create the tables.~~
- ~~Implements option 3 outlined in the issue by detecting if all of the expected tables exist, and if any are missing, runs `Migrator::up` to create the tables.~~
- Implements option 3 outlined in the issue. It detects if the database contains no tables, and if so runs `Migrator::up` to create the tables.

### Testing

This resolves the issue on `main` and also resolves the ("Unresolved") problem noted in #29

### DIFF
--- a/glados-monitor/src/main.rs
+++ b/glados-monitor/src/main.rs
@@ -36,31 +36,6 @@ async fn main() -> Result<(), DbErr> {
         "database connection established"
     );
 
-    let manager = SchemaManager::new(&conn);
-    let mut db_exists = true;
-    for t in [
-        "content_audit",
-        "content_id",
-        "content_key",
-        "node",
-        "record",
-    ] {
-        if !manager
-            .has_table(t)
-            .await
-            .expect("could not query database")
-        {
-            db_exists = false;
-        };
-    }
-
-    if !db_exists {
-        info!("creating new database tables");
-        Migrator::up(&conn, None)
-            .await
-            .expect("could not create new database tables");
-    }
-
     if cli.migrate {
         info!("running database migrations");
         Migrator::up(&conn, None).await.unwrap();
--- a/glados-monitor/src/main.rs
+++ b/glados-monitor/src/main.rs
@@ -36,6 +36,19 @@ async fn main() -> Result<(), DbErr> {
         "database connection established"
     );
 
+    let manager = SchemaManager::new(&conn);
+    let mut db_exists = true;
+    for t in ["content_audit", "content_id", "content_key", "node", "record"] {
+        if !manager.has_table(t).await.expect("could not query database") {
+            db_exists = false;
+        };
+    }
+
+    if !db_exists {
+        info!("creating new database tables");
+        Migrator::up(&conn, None).await.expect("coult not create new database tables");
+    }
+
     if cli.migrate {
         info!("running database migrations");
         Migrator::up(&conn, None).await.unwrap();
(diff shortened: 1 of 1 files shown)


pipermerriam:
can you rebase from master and resolve the conflict?  This should also trigger a CI run now as well.

glados-monitor/src/main.rs
```diff
@@ -41,6 +42,19 @@ async fn main() {
...
+            db_exists = false;
+        };
+    }
+
+    if !db_exists {
+        info!("creating new database tables");
+        Migrator::up(&conn, None).await.expect("could not create new database tables");
+    }
+
     if cli.migrate {
```
pipermerriam, line 58: I'm not sure I'm onboard with this approach (but I'm willing to be convinced).  The primary reason is that it unconditionally does database modifications which could have unexpected behavior in the event that the database is in a weird state.

It seems that we could do something like modifying this condition to be:

```rust
if (cli.migrate | database_is_in_memory_sqlite) {
    ...  // run migrations
}
```

thoughts?

glados-monitor/src/main.rs
```diff
@@ -41,6 +42,19 @@ async fn main() {
         "database connection established"
     );
 
+    let manager = SchemaManager::new(&conn);
+    let mut db_exists = true;
+    for t in ["content_audit", "content_id", "content_key", "node", "record"] {
```
pipermerriam, line 47: another reason I'm not sold on this approach is that this code will end up "ageing" if we add new database tables, and we might forget to update this list of required tables.  It'd be nice to use a solution that just focuses on letting the migration engine do this job and assume the migrations will do the correct thing when run.

glados-monitor/src/main.rs
perama-v: As @mrferris noted in #30, a file-based db with no contents also requires the db tables to be created. So an in-memory check alone doesn't handle that.

glados-monitor/src/main.rs
perama-v: Agree here. 

I've come up with a solution (f248981) that finds all tables in the database. If there are none, then it proceeds to make the necessary tables (via `Migrator::up()`). 

This works for in-memory (there are always no tables) and for persistent 
- If the database does not exist, the connection fails
- If the database exists, but is empty, `Migrator::up()` is called
- If the database exists, but is not empty, no migration is called (unless explicitly called with the `--migrate` flag)

perama-v:
I've tested the new `no_tables()` function for sqlite (memory and file-based). For postgres and mysql I have put queries that I believe will work along with a warn log that they are untested. Are these db's intended to be supported, or are they just a feature because they're built in to sea_orm?

pipermerriam:
No intention to ever need-or-support mysql.  We want to continue support for sqlite and postgres (local dev and deployments)

pipermerriam:
I think my preference is to keep this simple and simply auto-migrate when the database is in-memory and require manual migration for the other cases.

- in-memory: always performs migration
- on-disk sqlite: requires manual migration via `--migrate`
- postgres: requires manual migration via `--migrate`

I think that this gives us all the functionality needed to run the app locally and in production and side-steps needing to try and detect things.

perama-v:
Ok, back to the simple check for in-memory otherwise use --migrate flag manually (option 4).
- in-memory always triggers creation via `Migrator::up()`
- A database with no tables will panic and no `--migrate` flag will panic. I've included an explanation for this in the `SETUP_GUIDE.md` in #44.
- A database with no tables will panic and no `--migrate` flag will crate the tables via `Migrator::up()`

I will squash this if you agree

perama-v:
closes #31

PR #34, opened 2023-01-25
Use ethportal api crate

pipermerriam:
convert to use the new `ethportal-api` crate types.


glados-monitor/src/lib.rs
```diff
@@ -124,17 +132,25 @@ async fn retrieve_new_blocks(
...
+    name: &str,
+    conn: &DatabaseConnection,
+) where
+    Vec<u8>: From<&'b T>,
+{
+    let encoded: Vec<u8> = key.into();
+
     debug!(
-        content.key=key.hex_encode(),
+        content.key=format!("{:x?}", encoded),
```
perama-v, line 145: In testing this, I noticed that this logs a byte array (previously was logging the full hex encoded string)

Changing to the below was more readable.
```rust 
    let bytes: Vec<u8> = key.into();
    let encoded: String = hex::encode(bytes);

    debug!(
        content.key=format!("0x{encoded}"),
```
The `OverlayContentKey` in `ethportal-api` might benefit from an `into_hex_encoded()` method for this scenario.

entity/src/contentkey.rs
```diff
@@ -51,20 +55,28 @@ pub async fn get_or_create(
...
     content_key.insert(conn).await
 }
 
-pub async fn get(content_key: &impl ContentKey, conn: &DatabaseConnection) -> Option<Model> {
+pub async fn get<'b, T: OverlayContentKey>(
+    content_key: &'b T,
+    conn: &DatabaseConnection,
+) -> Option<Model>
+where
+    Vec<u8>: From<&'b T>,
```
mrferris, line 75:
```suggestion
pub async fn get(
    content_key: &impl OverlayContentKey
    conn: &DatabaseConnection,
) -> Option<Model>
```

mrferris (commented):
Having trin types in glados is awesome 👍 

Main feedback is that I prefer the syntax: `some_func(content_key: &impl OverlayContentKey)` 
over the more complicated:
 `some_func<'b, T: OverlayContentKey>(content_key: &'b T) where ...` syntax.

Let me know if there's a reason for having done it that way that I'm missing.

entity/src/contentaudit.rs
```diff
@@ -63,9 +63,16 @@ pub async fn create(
         .expect("Error inserting new content_audit")
 }
 
-pub async fn get_audits(content_key: &impl ContentKey, conn: &DatabaseConnection) -> Vec<Model> {
+pub async fn get_audits<'b, T: OverlayContentKey>(
+    content_key: &'b T,
+    conn: &DatabaseConnection,
+) -> Vec<Model>
+where
+    Vec<u8>: From<&'b T>,
```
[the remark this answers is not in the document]
pipermerriam: When I make this change the compiler gets mad about a bunch of things, so I think these two are not simply interchangeable.

glados-monitor/src/lib.rs
pipermerriam: I'm gonna dig at this in a subsequent PR.  Tracking here in https://github.com/pipermerriam/glados/issues/46

PR #42, opened 2023-01-26
add configuration guide

perama-v:
A guide to follow for setting up the various components. The examples are for a specific environment/configuration. 

They likely are beneficial even if they don't match the users setup, to see how the various binaries interact (transports, databases etc)


SETUP_GUIDE.md
```diff
@@ -0,0 +1,61 @@
...
+    - `glados-audit`
+    - `glados-web`
+- "`machine-b`"(with display)
+    - Browser listening to `glados-web`
+
+All commands are issued on `machine-a` unless otherwise stated.
+
+Important notes:
+- If `glados-monitor`, `glados-audit` and `glados-web` are in `sqlite::memory:`-mode they won't be able to share a database.
+- The database used is ephemeral (for testing) and will be removed if the machine reboots.
```
pipermerriam, line 20:
```suggestion
- If `glados-monitor`, `glados-audit` and `glados-web` are in `sqlite::memory:`-mode they won't be able to share a database. In-memory databases are ephemeral and only persist as long as the process is running.
```

PR #44, opened 2023-01-26
add clarification and description to guide

perama-v:
### Description

Adds a few additional descriptions to `SETUP_GUIDE.md`

### Changes
- Clarify that the use of `/tmp` for db files is vulnerable on reboot
- Remove the requirement on pr 31 (db check for tables to determine init) and show use of `--migrate` flag for new empty db creation
- Descriptions for what source/dest each binary has for data

### DIFF
--- a/SETUP_GUIDE.md
+++ b/SETUP_GUIDE.md
@@ -10,21 +10,22 @@ This example uses the following:
     - `glados-monitor`
     - `glados-audit`
     - `glados-web`
-- "`machine-b`"(with display)
+- "`machine-b`" (with display)
     - Browser listening to `glados-web`
 
 All commands are issued on `machine-a` unless otherwise stated.
 
 Important notes:
 - If `glados-monitor`, `glados-audit` and `glados-web` are in `sqlite::memory:`-mode they won't be able to share a database. In-memory databases are ephemeral and only persist as long as the process is running.
+- `glados-audit` does not support `http` communication (must start `trin` with `--web3-transport ipc`)
 
 Start Ethereum execution client (not covered here).
 
 Make an empty database file for glados.
 ```command
-$ touch /tmp/glados-ephemeral.sqlite
+$ touch /path/to/database.sqlite
 ```
-Start trin (requires https://github.com/ethereum/trin/pull/510), see repo for latest docs.
+Start trin (see [docs](https://github.com/ethereum/trin/blob/master/docs/ubuntu_guide.md)
 ```command
 ~/trin$ RUST_LOG=debug cargo run -p trin -- \
     --no-stun \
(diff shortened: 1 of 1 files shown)


SETUP_GUIDE.md
```diff
@@ -10,21 +10,24 @@ This example uses the following:
...
 - If `glados-monitor`, `glados-audit` and `glados-web` are in `sqlite::memory:`-mode they won't be able to share a database. In-memory databases are ephemeral and only persist as long as the process is running.
+- If a file based database is used and set to the `/tmp` directory
+(`sqlite:////tmp/glados-ephemeral.sqlite`), the database will be removed upon reboot.
+- `glados-audit` does not support `http` communication (must start `trin` with `--web3-transport ipc`)
 
 Start Ethereum execution client (not covered here).
 
 Make an empty database file for glados.
 ```command
 $ touch /tmp/glados-ephemeral.sqlite
```
pipermerriam, line 28: One thing that I tend to do when writing guides like this is to avoid using actual file paths and instead do things like this:

```
$ touch /path/to/database.sqlite
```

This does two things that I like:

1. It doesn't cement in a "concrete" path that may or may not work well on everyone's machine.  For example, if the user happens to already have a previous database file at this path, they may not notice it in their second pass through because they simply copy/paste commands.
2. It makes it clear that the path is under their control and that they can use any path they want.

In this specific case, it eliminates the need to make any disclaimers about using temporary files, because the use will be in control of setting the path themselves.

pipermerriam (approved):
Thank you for writing these guides up.  I really appreciate it and hope that my feedback doesn't feel too nitpicky.

pipermerriam:
I'm gonna leave this and merge next week in case you care/want to make any adjustments over the weekend.

SETUP_GUIDE.md
perama-v: Nitpicky feedback is welcome, don't hold back

perama-v: Done (paths changed)

njgheorghita:
+1 for adding detail on the `--migrate` flag to the docs 🥳

njgheorghita:
The docs indicate that the user **must** be running an execution client locally, but I've been using an `infura` url, and it's been working just fine. Just a thought, but if the docs specify that the user is "required" to run a local execution node, as opposed to infura, this could present a bit of unnecessary friction for somebody who just wants to experiment with glados.

perama-v:
> The docs indicate that the user **must** be running an execution client locally

I envisioned that other headings could exist for different setups. A user can quickly find the right heading and get going 

## Local testing environment
This PR
## Testing using infura 
Separate PR
## [some other common config]
Separate PR

Issue #35, opened 2023-01-25
Add meta/context information to the content-key/content-id information in the database.


pipermerriam:
We need to add some new information to our database that lets us have a richer view into what each content-key/content-id represents.

The new tables should loosely be something like:

```
execution_header
---
block_number: uint64
block_hash: [u8, 32]
content_id: foreign_key_to(content_id)

execution_body
---
block_number: uint64
block_hash: [u8, 32]
content_id: foreign_key_to(content_id)
```

execution_receipts
---
block_number: uint64
block_hash: [u8, 32]
content_id: foreign_key_to(content_id)
```
```

The `glados-monitor` process needs to be updated to populate these tables with information.

pipermerriam:
The database schema we pick for this should be intentionally friendly to the types of queries we'll need to do in https://github.com/pipermerriam/glados/issues/36

perama-v:
I'll look into this one

pipermerriam:
I started poking this gently yesterday: https://github.com/pipermerriam/glados/tree/piper/setup-schema-for-execution-history-data

Feel free to run with this however you want, either building off what I did in that branch or just doing your own thing.

PR #45, opened 2023-01-27
Setup schema for execution history data

perama-v:
### Description

Adds the ability of `glados-monitor` to store headers, bodies and receipts as described in #35. Each as a separate table.

This allows for richer context for audit results. For example: `header for block 16566152 (hash: 0xc4c3…b26f)` as shown below:
```sh
content.key="0x0192588055ccd2419bbbaed661420008f0a5713416e1cd72dd796e4562dc959e6c" block="body for block 16581206 (hash: 0x9258…9e6c)" audit.pass=false
```

### Changes

- `ExecutionHeader`, `ExecutionBody` and `ExecutionReceipts` Enums defined in separate migration files.
- Auto-implemented (`sea-orm-cli`) entities in `./entities/src` for the above.
- `DATABASE.md` readme for high level overview of databases.
- Logic
    1. Monitor: When a new block is retrieved, the number and hash are noted
    2. Monitor: `store_block_keys()` is called 
        a. Stores the content_key (header, body, receipt)  (as before)
        b. Stores the block number and hash in each of the new tables (new).
    3. Audit: For each content_key job received
        a. Perform the audit (as before)
        b. Look up the content_key in each of the new tables 
        c. Display the block number and block hash of the audited key

### Tasks
- [x] Integrate tables into `glados-monitor::retrieve_new_blocks()`. For every ContentKey that is being recorded with a ContentId, also record an entry in the new tables (mapping the ContentKey to a blockhash and blocknumber). 
- [x] ~~Address TODOs around enabling OverlayContentKey as reference (e.g., by exposing internal data using a new method)~~ Using OverlayContentKey bytes is handled in a separate PR (# 48).
- [x] Decide if block_hash should be removed from the new tables. Given that a content_key inherently contains this information. Outcome: Leave in for now.
- [x] Testing
- [x] Tidy up commits


entity/src/executionbody.rs
```diff
@@ -0,0 +1,70 @@
...
+impl Related<super::contentid::Entity> for Entity {
+    fn to() -> RelationDef {
+        Relation::ContentId.def()
+    }
+}
+
+impl ActiveModelBehavior for ActiveModel {}
+
+pub async fn create(
+    content_key_raw: &impl ContentKey,
```
pipermerriam, line 43: it seems we should be able to derive the content key from the block_number/hash values?

entity/src/executionbody.rs
```diff
@@ -0,0 +1,70 @@
...
+
+impl Related<super::contentid::Entity> for Entity {
+    fn to() -> RelationDef {
+        Relation::ContentId.def()
+    }
+}
+
+impl ActiveModelBehavior for ActiveModel {}
+
+pub async fn create(
```
pipermerriam, line 42: For naming conventions, I'd name this `get_or_create` since it looks up the existing entry *if* it exists, and otherwise creates it.

glados-monitor/src/lib.rs
```diff
@@ -109,22 +112,80 @@ async fn retrieve_new_blocks(
             "received block",
         );
 
+        let num = block_number_to_retrieve;
         let hash = H256::from_slice(block_hash.as_bytes());
 
         let header = BlockHeaderContentKey { hash };
-        let body = BlockHeaderContentKey { hash };
```
pipermerriam, line 115: :heart:

glados-monitor/src/lib.rs
```diff
@@ -143,6 +204,12 @@ async fn store_content_key<T: ContentKey>(key: &T, name: &str, conn: &DatabaseCo
     }
 }
 
+enum Component {
```
pipermerriam, line 207: I've just merged the `ethportal-api` integration which should clean this up.

pipermerriam (commented):
wonderful, I'll give it another look when you mark it actually ready for review (sorry for the early-review)

perama-v:
I was going to ping you for an early look anyway, thanks

entity/src/contentkey.rs
```diff
@@ -36,14 +36,15 @@ pub enum Relation {
...
     conn: &DatabaseConnection,
-) -> Result<Model, DbErr>
-where
-    Vec<u8>: From<&'b T>,
-{
-    let encoded: Vec<u8> = content_key_raw.into();
+) -> Result<Model, DbErr> {
+    // The passing of &OverlayContentKey is currently limited (requires lifetimes).
+    // This is a temporary fix / reminder. Likely solutions are that OverlayContentKey should:
+    // 1. Have a method `fn bytes(&self) -> [u8; u32]` that can replace `clone()` here.
```
perama-v, line 45: See https://github.com/ethereum/trin/issues/518

perama-v:
Includes quick-fix for https://github.com/pipermerriam/glados/issues/50 by adding an "0x".

entity/src/executionbody.rs
perama-v: We already have the content_key at the time of calling this function. So deriving it is possible from the block hash but its already done.

perama-v: Oh I think I see what you mean. We store the block_hash in the three new tables, but this is duplicate information. So the new tables could theoretically just have the block number.

perama-v:
Ok this is now functional. Following an audit, `glados-audit` looks up the new block metadata and includes that in
the logged audit result. 

Closes #35

entity/src/executionbody.rs
```diff
@@ -0,0 +1,90 @@
...
+}
+
+/// Returns the an execution body monitor item if present.
+///
+/// Used during auditing. Returning none indicates a failed audit.
+///
+/// The foreign id is the database-assigned id for a content_key
+/// (not the portal network content_id).
+pub async fn get(
+    // The database-assigned id for a content key.
```
pipermerriam, line 80: The docs and variable names seem to be referring to content *keys* where-as the schema and database relations are all about content *id*.

This has me questioning whether all of the header/body/receipts tables should be pointed at the *ContentKey* table instead.  Thoughts?

glados-audit/src/lib.rs
```diff
@@ -91,10 +92,81 @@ async fn perform_content_audits(
...
+                debug!(
+                    content.key=?content_key,
+                    block = b.to_string(),
+                    audit.pass=?audit_result,
+                );
+            }
+            Ok(None) => {
+                warn!("Block metadata absent for key: {content_key}, with audit result: {audit_result}");
+            }
+            Err(e) => warn!("Problem getting block metadata {}", e),
```
pipermerriam, line 111: can this get updated to use structured logging for the error data, something simple like `error=e`

glados-monitor/src/lib.rs
```diff
@@ -111,54 +112,97 @@ async fn retrieve_new_blocks(
...
+    store_content_key(&header, name, conn).await;
+
+    let name = "block-body";
+    match executionbody::get_or_create(&body, block_number, block_hash, conn).await {
+        Ok(_) => log_record_outcome(&body, name, DbOutcome::Success),
+        Err(e) => log_record_outcome(&body, name, DbOutcome::Fail(e)),
+    }
+    store_content_key(&body, name, conn).await;
+
+    let name = "block-receipts";
```
pipermerriam, line 159: *nitpick*

I'm not sure what the best practices are in rust (I come from a heavy python background).  I get a small "code smells" from the re-use of the `name` variable here.  Very small foot-gun.

pipermerriam (approved):
Let me know when you're ready for me to hit the merge button (or if you want additional review on any component of this).  There's at least one thing that I left a comment on that needs to be cleaned up (the use of a `contentkey.id` when a `contentid.id` should be used)

entity/src/executionbody.rs
perama-v: Ok I see what you mean. 

Tracking/discussion: https://github.com/pipermerriam/glados/issues/52

glados-audit/src/lib.rs
```diff
@@ -91,10 +92,81 @@ async fn perform_content_audits(
         let content = client.get_content(&content_key);
 
         let raw_data = content.raw;
+        let audit_result = raw_data.len() > 2;
+        let contentkey_db_id = contentkey::get(&content_key, &conn).await.unwrap().id;
+        contentaudit::create(contentkey_db_id, audit_result, &conn).await;
+
+        // Display audit result with block metadata.
+        match fetch_block_metadata(contentkey_db_id, &conn).await {
```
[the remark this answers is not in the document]
perama-v: Good pickup

glados-monitor/src/lib.rs
perama-v: Fixed with non-identical variable names

glados-audit/src/lib.rs
perama-v: Done. For some reason info!, debug! and error! have the same behaviour, but warn! doesn't like the use of `?`. 

Upgraded to errors and included structured logs.

glados-audit/src/lib.rs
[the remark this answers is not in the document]
perama-v: The latter is complete (have replaced table columns `content_id` with `content_key`)

entity/src/executionbody.rs
perama-v: I am going to leave the three new tables as containing:

- content_key (via foreign)
- block_hash (which technically is contained within the linked content_key)
- block_number

Rather than 
- content_key (via foreign)
- block_number

If later we want to cut the block_hash out to reduce db size we can do so then.

perama-v:
Ok ready from my perspective

perama-v:
Is it ok for additional commits squashed like this?

Or would you prefer them pushed to make reviewing new changes easier. Then later I squash them after final approval (Adding an extra loop)

pipermerriam:
Not overly concerned with our git history.  I'll tend to squash during merge but we're in "move fast and break things" mode right now with this codebase so there's not a high need for consistent commit hygiene.

PR #48, opened 2023-01-30
Use bytes() method for OverlayContentKey

### LINKED ISSUE #47: OverlayContentKey use does not include selector
## Description

A `T: OverlayContentKey` passed into a function does not have a way to get the actual content_key (e.g., the bytes including the selector).  

Use of `.into()`, which is the counterpart to `From` and has the following implementation.
```rs
// trin/ethportal-api/src/types/content_key.rs
impl From<HistoryContentKey> for Vec<u8> {
    fn from(val: HistoryContentKey) -> Self {
        val.as_ssz_bytes()
    }
}
```
This gets the bytes of the enum, using the derived `Encode`
```rs
/// A content key in the history overlay network.
#[derive(Clone, Debug, Decode, Encode, Eq, PartialEq)]
#[ssz(enum_behaviour = "union")]
pub enum HistoryContentKey {
```
Which does not include the selector.

perama-v:
## Description

- Explicitly uses a new `.to_bytes()` method common to all `OverlayContentKey`s.
    - Introduced in: https://github.com/ethereum/trin/pull/519
- Bumps to latest `ethportal-api = "0.1.6"`

## Usage
The general pattern this introduces is the ability to access the bytes of a content key. 
```rs
fn use_key<T: OverlayContentKey>(key: &T) {
    let content_key = key.to_bytes(); // content_key: Vec<u8>
}
```

## Testing
- [x] Audits run as expected


perama-v:
Moving this to draft because the method is subject to change. It is likely to:
- Return `Vec<u8>` not `[u8, 33]`.
- Be named `.to_bytes()`

perama-v:
Ready now that `ethportal-api` `0.1.6` is published.

pipermerriam (approved):

PR #49, opened 2023-01-30
use anyhow for error handling

perama-v:
### Description

Use `anyhow` for error handling. Predominant benefit is quick and easy handling, especially in the context of
returning multiple error types within a single function. `anyhow` preserves error types, bubbling them up. 

The errors encountered in major code paths (such as loop blocks following the chain tip) are all unpacked and converted to
logs. This should ideally result in very few panics.

### Patterns

These are the rules that I used:

- Import `anyhow::Result` and use that everywhere instead of `std::Result`
- No use of `.unwrap()`, just `?` on the thing and `Result<()>` in signature.
    - Except in `glados-web/src/routes.rs`, axum did not work well with anyhow for some reason at first glance. `unwrap()`s are left there for now.
- Don't put error types in the signature like `Result<(), SpecificError>`. This allows multiple error types in one function.
- Where no value is returned, use 
    - Function signature is `Result<()>`
    - End function with `Ok(())`
- Where a function final expression is a call to another function with a result,
first allow anyhow to deal with the error (`?`), then wrap the success case: (in `Ok()`)
    - Pre `fallible_last_thing().unwrap()`
    - Post `Ok(fallible_last_thing()?)`
- Minimise `.expect("Panic message")`, instead if a message to be included alonside an error, use `anyhow::Context`. This keeps the error message and the new message 
    - `fallible_call().with_context(|| "lazy custom thing, even function calls" )?`    
    - This is useful for file opening errors, where you want to include the path that failed, but also include the actual error kind.
    - `let f = fs::read(path).with_context(|| format!("Failed to read {}", path.display()))?`
    - Note the `||` closure
- Use `bail!()` if there is a need to return with an error and a custom message
    - E.g., `let Some(t) = thing_or_none { else bail!("No thing, returning this error msg")}`

### DIFF
--- a/glados-core/src/jsonrpc.rs
+++ b/glados-core/src/jsonrpc.rs
@@ -6,6 +6,7 @@ use uds_windows::UnixStream;
 use std::path::Path;
 use std::str::FromStr;
 
+use anyhow::{bail, Context, Result};
 use discv5::enr::CombinedKey;
 use ethereum_types::{H256, U256};
 use ethportal_api::types::content_key::OverlayContentKey;
@@ -40,14 +41,14 @@ fn build_request<'a>(
 }
 
 pub trait TryClone {
-    fn try_clone(&self) -> std::io::Result<Self>
+    fn try_clone(&self) -> Result<Self>
     where
         Self: Sized;
 }
 
 impl TryClone for UnixStream {
-    fn try_clone(&self) -> std::io::Result<Self> {
-        UnixStream::try_clone(self)
+    fn try_clone(&self) -> Result<Self> {
+        Ok(UnixStream::try_clone(self)?)
     }
 }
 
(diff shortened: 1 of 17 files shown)


glados-core/src/jsonrpc.rs
```diff
@@ -60,10 +61,10 @@ where
 }
 
 impl PortalClient<UnixStream> {
-    pub fn from_ipc(path: &Path) -> std::io::Result<Self> {
-        // TODO: a nice error if this file does not exist
+    pub fn from_ipc(path: &Path) -> Result<Self> {
         Ok(Self {
-            stream: UnixStream::connect(path)?,
+            stream: UnixStream::connect(path)
+                .with_context(|| format!("Could not open ipc file {}", path.display()))?,
```
perama-v, line 67: Here is a good example of `.with_context()`

pipermerriam (approved):

pipermerriam:
Thanks for this cleanup.

PR #56, opened 2023-02-09
add welcome message for glados-web

perama-v:
### Description

There was no indication of success/failure upon `glados-web` startup

### Changed

Added welcome message `info!()` log:
```sh
...INFO  glados_web] Serving glados-web at <IP>:<Port>
```
Which reads the `SocketAddr` variable.

### Related

- Adds `tracing` and `env_logger` crates

### DIFF
--- a/glados-web/src/lib.rs
+++ b/glados-web/src/lib.rs
@@ -1,5 +1,5 @@
-use std::path::Path;
 use std::sync::Arc;
+use std::{net::SocketAddr, path::Path};
 
 use anyhow::{bail, Result};
 use axum::{
@@ -8,6 +8,7 @@ use axum::{
     Router,
 };
 use tower_http::services::ServeDir;
+use tracing::info;
 
 pub mod cli;
 pub mod routes;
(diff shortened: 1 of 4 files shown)


glados-web/src/lib.rs
```diff
@@ -39,8 +42,10 @@ pub async fn run_glados_web(config: Arc<State>) -> Result<()> {
         .fallback_service(serve_dir)
         .layer(Extension(config));
 
-    // run it with hyper on localhost:3000
-    Ok(axum::Server::bind(&"0.0.0.0:3001".parse()?)
+    let socket: SocketAddr = SOCKET.parse()?;
+    let port = socket.port();
+    info!("Serving glados-web at 127.0.0.1:{port}");
```
pipermerriam, line 47: This string looks like it'll probably end up being incorrect at some point.  Can we not pull a pretty-printed version of the IP address from the socket?

perama-v: Ok, I've change it to use the full socket
```rs
info!("Serving glados-web at {}", socket);
```

PR #57, opened 2023-02-09
Add metadata to glados web

### LINKED ISSUE #53: Add block metadata to web view
The web application should be updated to include the new block metadata in any contentkey/id displays so that when we view a content key/id in the web application we can know if it is a header/body/receipts/accumulator

perama-v:
### Description

Add information about what a key represents in `glados-web` UI. 

- Fixes #53.

### Depends on

- This PR is rebased on #59 

### Changes
- Uses ethportal-api types in glados-web
- Recreates `HistoryContentKey` from `content_key` bytes that `glados-web` pulls from db.
- Uses `.to_string()` method to display content kind (header, body, receipts, accumulator)
- Performs a lookup of a block number if the content key is not for an accumulator 0x00, 0x01, 0x02.

### Effect

Navigation is unchanged and is as follows:
- Main page (list of content_ids)
    - content_id page (link to content_key)
        - content_key page: (see below)
        
In addition to the audit results, the browser UI now shows:
```
Database ID: 9989
Content Key: 0x0031b4baa7fcd9b087c244eebfbeb8091dac37a44f719df5fc6c848c76ccd6cfdb
Content Id: 0xffe59a4b228fc30120f706fbb8da01a8a217e7439af019ff3268b54ae612f1c6
Kind: BlockHeader { block_hash: 0x31b4..cfdb }
Block number: 16600489
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1728,6 +1728,7 @@ dependencies = [
  "entity",
  "env_logger",
  "ethereum-types 0.14.0",
+ "ethportal-api",
  "glados-core",
  "hex",
  "migration",
--- a/glados-audit/src/lib.rs
+++ b/glados-audit/src/lib.rs
@@ -43,7 +43,10 @@ where
     let mut interval = interval(Duration::from_secs(AUDIT_PERIOD_SECONDS));
     loop {
         interval.tick().await;
-
+        if tx.is_closed() {
+            error!("Channel is closed.");
+            panic!();
+        }
         // Lookup a content key to be audited
         let content_key_db_entries = match contentkey::Entity::find()
             .order_by_desc(contentkey::Column::CreatedAt)
@@ -97,7 +100,7 @@ async fn perform_content_audits(
 where
     Vec<u8>: From<HistoryContentKey>,
 {
-    let mut client = PortalClient::from_ipc(&ipc_path)?;
+    let mut client = PortalClient::from_ipc(&ipc_path).expect("Could not connect to portal node.");
 
     while let Some(content_key) = rx.recv().await {
         let content_key_str = format!("0x{}", hex::encode::<Vec<u8>>(content_key.clone().into()));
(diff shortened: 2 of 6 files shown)


migration/src/m20221114_143914_create_content_id_key_and_audit.rs
```diff
@@ -47,27 +19,25 @@ impl MigrationTrait for Migration {
...
-                    .col(ColumnDef::new(ContentKey::ContentId).integer().not_null())
-                    //.index(Index::create().name("idx-contentkey-content_id").col(ContentKey::ContentId))
-                    .foreign_key(
-                        ForeignKey::create()
-                            .name("FK_conent_key_content_id")
-                            .from(ContentKey::Table, ContentKey::ContentId)
-                            .to(ContentId::Table, ContentId::Id)
-                            .on_delete(ForeignKeyAction::Cascade)
-                            .on_update(ForeignKeyAction::Cascade),
+                    .col(
```
pipermerriam, line 22: I don't think it is viable to muck with old migration files (since they will have already been executed in some environments).  We will need new migrations that correctly modify/drop existing tables, otherwise these are going to fail when I try to run them in the production environment.

pipermerriam (commented):
I think I'd like to ask that you test these with a real (postgres) database that is non-empty and which was migrated from master to ensure that the migrations which modify the various tables work as expected.

migration/src/m20221114_143914_create_content_id_key_and_audit.rs
perama-v: Ok that's fair. Just a note - this PR is a rebase on top of the DB changes ("requires PR 55"), so maybe the discussion can happen there

perama-v: This PR is now completely isolated from / unrelated to database changes

perama-v:
Ready to go once #59 bugfix is merged.

perama-v:
My pleasure. Rebased on `main` now

PR #59, opened 2023-02-10
Change block_number i64 -> i32

### LINKED ISSUE #58: New block_number feature clash with postgres
### Description

When main branch is run with a (new/empty) Postgres backend the following error occurs: 
```
Query Error: error occurred while decoding column "block_number": mismatched types; 
Rust type `core::option::Option<i64>` (as SQL type `INT8`) is not compatible with SQL type `INT4`
```
### Origin
This was introduced by PR #45, which was only tested using sqlite. Root cause TBD. 

### Full logged error
```
Failed to create database record 
content.key="0x004be441720f239cdc201bcafc41a29d86f5f4056005d0af29176ce0e19ade2c33" 
content.kind="header_metadata" 
err=Query Error: error occurred while decoding column "block_number": mismatched types; 
Rust type `core::option::Option<i64>` (as SQL type `INT8`) is not compatible with SQL type `INT4`
```

### Solution
- Possibly downgrade block_number to be a 32 bit not 64 bit.

perama-v:
### Description

- Fixes #58. 

### Changes

Makes `block_number` an `i32`. 

### Testing

- Tested on fresh postgres
- Tested on fresh sqlite

### DIFF
--- a/entity/src/executionbody.rs
+++ b/entity/src/executionbody.rs
@@ -14,7 +14,7 @@ pub struct Model {
     #[sea_orm(unique)]
     pub content_key: i32,
     #[sea_orm(unique)]
-    pub block_number: i64,
+    pub block_number: i32,
     #[sea_orm(unique)]
     pub block_hash: Vec<u8>,
 }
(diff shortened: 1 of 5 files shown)


mrferris (approved):
By my calculations, 32 bits signed should hold the block number without overflow for another 900 years. By that time, Postgres will probably support unsigned integers, which will buy us another 900 years 👍

PR #69, opened 2023-02-24
Simplified schema

### LINKED ISSUE #68: Make auditing compatible with different sub-protocols
## Description

The auditing is specific to the History network. If it were to be made agnostic to sub-protocols then this will save work later on.
The cause of the current limitation is that we:
1. ✅ (`glados-monitor`) Follow a Portal node and record content keys/ids/metadata  in a glados-db
2. ✅ (`glados-audit`) Employ different strategies to decide on what glados-db content to audit.
3. (`glados-audit`) presume all content in glados-db are from the History sub-protocol and send `HistoryContentKey`s in the mpsc audit channel. ❌ Would not handle other sub-protocols.
4. (`glados-audit`) For each task in the audit channel, perform a `portal_historyRecursiveFindContent` request to a Portal Node.  ❌  Would not handle other sub-protocols.

## Solution
In three parts:
### Record the sub-protocol in the glados-db. 
Currently we do this with the execution metadata tables, for header, body & receipts but not EpochAccumulator. Options are:
A. Treat these tables as sub-protocol identifiers. Would need to add a table for EpochAccumulator items, and any items added for different sub-protocols
B. Add a new table for sub-protocol

Option B seems cleaner because in the event of a second sub-protocol, you do not need to check multiple tables to find matches.

### Make mpsc channel more broad

The channel currently handles `HistoryContentKey`. It could send database IDs in the channel. 

### Lookup sub-protocols at audit time

The audit task can operate as follows:
- 1. Look up what sub-protocol the item is from
- 2. Call the appropriate `portal_*RecursiveFindContent`
- 3. Log the appropriate metadata (based on the sub-protocol)

## Considerations

Any key entered into the glados-db may clash with an identical key on a different sub-protocol. To handle this,
keys should AFAICT have a many-to-one relationship with a sub-protocol identifier. 

- Decide on a key to audit
    - Look up what sub-protocol it is from
    - Audit on that sub-protocol (E.g., `portal_historyRecursiveFindContent`)
    - Record audit result for that content
- Glados monitor creates a second key that is identical (to the above) but on a different subprotocol
    - Need to store the key, but in a way so as to not inherit the audit from the similar key 
        - Perhaps by storing the sub-protocol along with the audit date/result data.
    
Not yet sure of the best way to a organise a sub-protocol table / foreign key to achieve this.

## Alternatives

Maybe there are other better solutions. 
- ? Separate database for sub-protocols

perama-v:
### Description

Redesign of the glados database. Now consists of:
- Content table 
- Metadata table
- Audit table

Breaking change to the database to include additional data. Implemented as a database wipe rather than a migration.
- Resolves #68
 
### Additional changes
As the database is being wiped this was an opportunity to improve the metadata tables as well.
- Combine the metadata tables into one
- Remove the block hash from the block metadata table (is obtainable given a content key)

### Checklist
- [x] tidy commits
- [x] `entity/test.rs`
- [x] Blank SQLite testing.
    - Glados monitor
    - Glados audit
    - Glados web
    - Browser UI navigation
- [x] Blank PostgreSQL testing.
    - Glados monitor
    - Glados audit
    - Glados web
    - Browser UI navigation


pipermerriam (commented):
Am I correct that once this is merged, we'll want to fully wipe the production database (delete the whole database) and recreate it, since it appears that we're doing a full reset of the migration files/history?

perama-v:
> Am I correct that once this is merged, we'll want to fully wipe the production database (delete the whole database) and recreate it, since it appears that we're doing a full reset of the migration files/history?

Yes, that is correct

entity/src/content.rs
```diff
@@ -0,0 +1,102 @@
...
+    History = 0,
+    State = 1,
+}
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq)]
+#[sea_orm(table_name = "content")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    pub id: i32,
+    #[sea_orm(unique)]
```
pipermerriam, line 21: This `unique` is suspect to me.  I haven't actually looked at the migration files yet but I can imagine this being confusing to someone reading the code.

pipermerriam:
I went ahead and rage-merged this so that I can start some work on top of it and go ahead and get it deployed to production for some exciting "live" testing.

PR #73, opened 2023-03-07
Remove unique constraint on contentaudit's contentkey column

mrferris:
The `content_audit` model mistakenly had a `unique` constraint on its `content_key` column, which will lead to errors when trying to re-audit a particular content key.


pipermerriam:
This will need a database migration to remove the constraint at the database level.  You'll need to generate a new migration file, and then drop the index (you will need to look at the previous migration files to find the name of the index that needs to be dropped)

perama-v:
Good find! I have also applied this change to the simplified schema PR
- #69

mrferris:
@perama-v Ok great, I'll close this then

PR #76, opened 2023-03-10
add record of selection strategy used to glados

### LINKED ISSUE #63: Audits do not record the originating strategy
### Description

> One suggestion for a future PR is that we include within the audit's DB record which selection strategy it was triggered by.

When an audit fails, it may be exclusively from a particular strategy. By incorporating the source strategy such correlations could be interrogated

### LINKED ISSUE #67: Add concurrency to auditing
### What is wrong

Currently, I believe our auditing is effectively single threaded, aka, we only ever hit our connected portal node (`trin`) with a single running content lookup.  Since content lookups can take a bit of time due to needing to traverse the network, we should probably run multiple of these concurrently.

### How can it be fixed

Change `glados-audit` to be able to run multiple lookups concurrently.

Add a new flag to `glados-audit` such as `--concurrency N` which sets the number of concurrent lookups that can be running at any given time.

perama-v:
### Description

- Fixes #63 
- The most convenient way to pass the strategy used to the auditor was to have a new thread for every audit strategy. This also fixes #67 to some extent (adds concurrency)


glados-audit/src/lib.rs
```diff
@@ -15,19 +15,27 @@ pub mod cli;
...
-    let (tx, rx) = mpsc::channel::<HistoryContentKey>(100);
     let strategies = vec![
         SelectionStrategy::Latest,
         SelectionStrategy::Random,
         SelectionStrategy::Failed,
         SelectionStrategy::OldestMissing,
     ];
     for strategy in strategies {
-        tokio::spawn(strategy.start_audit_selection_task(tx.clone(), conn.clone()));
+        let (tx, rx) = mpsc::channel::<HistoryContentKey>(100);
```
mrferris, line 25: Curious about your reasoning for using a separate channel for each strategy, rather than a single channel for all strategies like previously done

entity/src/content_audit.rs
```diff
@@ -13,6 +13,20 @@ pub enum AuditResult {
     Success = 1,
 }
 
+/// The audit strategy used for the given result
+///
+/// This enum shadows the glados-audit::selection::SelectionStrategy.
+/// SelectionStrategy itself cannot be used because it creates a cyclic
```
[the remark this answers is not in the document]
perama-v: Ok good idea

glados-audit/src/lib.rs
perama-v: I have changed it to be a single channel once more.

The original rationale was to that each receiver channel would know the strategy because it only handles that strategy (resolved by passing the strategy in the channel). 
Multiple threads was also in line with increasing concurrency for audit task completion (better resolved separately).

entity/src/content_audit.rs
```diff
@@ -13,6 +13,30 @@ pub enum AuditResult {
     Success = 1,
 }
 
+#[derive(Debug, Clone, Eq, PartialEq, EnumIter, DeriveActiveEnum)]
+#[sea_orm(rs_type = "i32", db_type = "Integer")]
+/// A strategy is responsible for generating audit tasks.
+///
+/// An audit task is a content key from the the glados database that
+/// is expected to be in a portal node.
```
mrferris, line 21:
```suggestion
/// Each strategy is responsible for selecting which content keys to begin audits for. 
```

PR #77, opened 2023-03-10
implement failed strategy

### LINKED ISSUE #62: Add additional audit strategies
### Description

At present the strategies are:
- `latest`
- `random`
- `failed` (prev `missing`)

New strategies suggested:

And then I think in the future we'd also want a `Missing` strategy that looks up data that has never been audited, but not necessarily the newest. Maybe `OldestMissing` that works its way forward from the beginning of the chain.

_Originally posted by @mrferris in https://github.com/pipermerriam/glados/pull/60#discussion_r1106250734_

perama-v:
## Issue addressed

- closes #62 by adding the failed strategy

## Proposed changes

Adds the `SelectionStrategy::Failed`

## Additional info

- [x] unit test for strategy

### DIFF
--- a/glados-audit/src/selection.rs
+++ b/glados-audit/src/selection.rs
@@ -14,7 +14,7 @@ use tracing::{debug, error, warn};
 
 use entity::{
     content::{self, Model},
-    content_audit,
+    content_audit::{self, AuditResult},
 };
 
 /// Interval between audit selections for a particular strategy.
@@ -55,7 +55,7 @@ impl SelectionStrategy {
         match self {
             SelectionStrategy::Latest => select_latest_content_for_audit(tx, conn).await,
             SelectionStrategy::Random => select_random_content_for_audit(tx, conn).await,
-            SelectionStrategy::Failed => warn!("Need to implement SelectionStrategy::Failed"),
+            SelectionStrategy::Failed => select_failed_content_for_audit(tx, conn).await,
             SelectionStrategy::OldestMissing => {
                 warn!("Need to implement SelectionStrategy::OldestMissing")
             }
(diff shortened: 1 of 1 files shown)


pipermerriam:
minor conflict to be cleaned up and I'll get this merged

pipermerriam (commented):
I think the querying in this strategy is going to need a bit more machinery (most of which I'm inclined to have tested to be sure we're actually doing things correctly).

First, I worry about race conditions in the "queue".  If I'm reading the code correctly, I think that this strategy will continue to select the same items if the rate the queue is being processed is slower than the rate this loop runs, which I believe results in the queue getting filled up with multiple copies of the same item.

Second, I am unsure if this is going to do the correct thing when a piece of content has multiple audits, some of which are failing and some of which are succeeding.  Are we sure that it will only pick the ones who's most recent audit has failed?  The code currently looks like it will filter out all successful audits, and then select the most recent failed one... which doesn't seem correct.

perama-v:
>  which I believe results in the queue getting filled up with multiple copies of the same item.

Agree with this observation. It will need to be addressed

> The code currently looks like it will filter out all successful audits, and then select the most recent failed one... which doesn't seem correct.

Agree. Understanding how to effectively execute a subquery for the solution using sea-orm remains elusive to me.

pipermerriam:
> Agree. Understanding how to effectively execute a subquery for the solution using sea-orm remains elusive to me.

Did you find this bit in the docs?

https://www.sea-ql.org/SeaORM/docs/advanced-query/subquery/

perama-v:
> Did you find this bit in the docs?

I did. I just haven't been able to put it together to do the sequence of things required for the strategy

PR #85, opened 2023-03-18
add http transport

### LINKED ISSUE #74: Support for HTTP(S) JSON-RPC interface in glados_audit
It would be nice to have support for HTTP for the JSON-RPC interface with the Portal client, instead of only IPC.

Fluffy does not support IPC.

perama-v:
## Issue addressed

- Closes #74 

## Proposed changes

Adds HTTP capability to `glados-audit`, where currently IPC exclusively supported. 
- The transport is selected with a new `--transport <ipc/http>`. 
- Transport configuration uses the existing `--ipc-path <path>` flag and a new `--http-url <url>` flag.
- New `AuditConfig` for holding config derived from parsed CLI arguments.

### CLI

This is the output from `cargo run -p glados-audit -- --help`

```console
Usage: glados-audit [OPTIONS] --transport <TRANSPORT>

Options:
  -d, --database-url <DATABASE_URL>  [default: sqlite::memory:]
  -i, --ipc-path <IPC_PATH>          
  -u, --http-url <HTTP_URL>          
  -t, --transport <TRANSPORT>        [possible values: ipc, http]
  -c, --concurrency <CONCURRENCY>    number of auditing threads [default: 4]
  -h, --help                         Print help information
  -V, --version                      Print version information
```

The transport flag is required. For the given transport variant (IPC/HTTP), if the corresponding path/url flag is not provided
an `error!()` is raised and the the audits do not start (`glados-audit` peacefully terminates)

### Configuration
New config built using the existing `clap`-based `Args` struct.
```rs
pub struct AuditConfig {
    /// Maximum amount of threads that audits will be performed with.
    pub concurrency: u8,
    /// For Glados-related data.
    pub database_url: String,
    /// For communication with a Portal Network node.
    pub transport: TransportConfig,
}
```
Where transport config is new and contains a transport-specific url/path.
```rs
pub enum TransportConfig {
    HTTP(Url),
    IPC(PathBuf),
}
```
### Client definition

The http method leverages the existing request code and introduces the change by converting the `PortalClient` struct into an enum. 
```rs
pub enum PortalClient {
    HTTP(HttpClient),
    IPC(IpcClient),
}
```
Where, when a new thread is created to complete an audit task, the aforementioned `TransportConfig` is used to create
a `let client = PortalClient::from_config()`. The request made using the existing method called `client.make_request(method, params)`.

The enum variants house the details required for sockets/http/ids as follows:
```
pub struct HttpClient {
    client: Client, // jsonrpsee async HTTP client
    // No request id as the client auto-increments.
}

pub struct IpcClient {
    stream: UnixStream, // existing IPC client
    request_id: u64,
}
```
### Response handling
Responses for HTTP (jsonrpsee crate) and IPC (`serde_json`-parsed locally-defined `JsonRPCResult`) are different. They are unified by an enum (`PortalResponse`) so that making a client request can be abstracted from the transport kind.
```rs
pub enum PortalResponse {
    ContentAbsent, // special case defined in portal network ("0x" == "no content")
    Regular(Value), // serde_json::Value
}
```
### Dependencies
The `jsonrpsee` crate is used for the client because it is used in `Trin`.

### Performance
The http client and ipc sockets are discarded after every request. This follows the concurrency model where tasks are
spawned in new threads, they create a connection and then conclude, terminating that connection. 

### Concurrency parsing
When the new `AuditConfig` struct is created a new check is introduced that reads available parallelism on the system via:
```rs
let parallelism = std::thread::available_parallelism()?.get();
```
A `warn!` is raised if the user passes a concurrency flag greater than this value. 
While this could be a separate PR, I have included it here because it is useful in testing to see what threads are available for the http client to use and is a minimal change. 

### Error handling in `glados-core::json_rpc`

All instances of error handling with `anyhow::Error` are replaced by custom `thiserror::Error` enum variants. This is to increase error clarity and information contextual available as `glados-core` is effectively used as a library by `glados-audit`.

### Additional notes

- [x] Investigate switching to IPC to jsonrpsee to simplify response handling. Result: Encountered a few issues not readily resolved. Simplification introduced by other means. Conclusion: Could be the subject of a different PR if desired.
- [x] Testing
    - [x] Flag usage 
    - [x] IPC with Trin functional
    - [x] HTTP with Trin functional 
- [x] Rebase on master once 81 is merged
     - Currently built on top of: #81 for testing.
- [x] Raised #92

### DIFF
--- a/glados-core/src/jsonrpc.rs
+++ b/glados-core/src/jsonrpc.rs
@@ -1,90 +1,125 @@
+use std::io::Write;
 #[cfg(unix)]
 use std::os::unix::net::UnixStream;
-use trin_utils::bytes::hex_decode;
-#[cfg(windows)]
-use uds_windows::UnixStream;
-
-use std::path::Path;
+use std::path::PathBuf;
 use std::str::FromStr;
 
-use anyhow::{bail, Context, Result};
 use discv5::enr::CombinedKey;
 use ethereum_types::{H256, U256};
 use ethportal_api::types::content_key::OverlayContentKey;
+use jsonrpc::{error::RpcError, Request};
+use jsonrpsee::{
+    core::{client::ClientT, params::ArrayParams},
+    http_client::{HttpClient, HttpClientBuilder},
+    rpc_params,
+};
 use serde::{Deserialize, Serialize};
-use serde_json::value::{to_raw_value, RawValue};
+use serde_json::{
+    json,
+    value::{to_raw_value, RawValue},
+    Value,
+};
 use thiserror::Error;
+use tracing::error;
+use trin_utils::bytes::hex_decode;
+#[cfg(windows)]
+use uds_windows::UnixStream;
+use url::Url;
 
 type Enr = discv5::enr::Enr<CombinedKey>;
 
-//
-// JSON RPC Client
-//
-fn build_request<'a>(
-    method: &'a str,
-    raw_params: &'a Option<Vec<Box<RawValue>>>,
+/// Configuration details for connection to a Portal network node.
+#[derive(Clone, Debug)]
+pub enum TransportConfig {
+    HTTP(Url),
+    IPC(PathBuf),
+}
+
+/// Details for a Connection to a Portal network node over different transports.
+pub enum PortalClient {
+    HTTP(HttpClientManager),
+    IPC(IpcClientManager),
+}
+
+/// HTTP-based transport for connecting to a Portal network node.
+pub struct HttpClientManager {
+    client: HttpClient,
+}
+
+/// IPC-based transport for connecting to a Portal network node.
+pub struct IpcClientManager {
+    stream: UnixStream,
     request_id: u64,
-) -> jsonrpc::Request<'a> {
-    match raw_params {
-        Some(val) => jsonrpc::Request {
-            method,
-            params: val,
-            id: serde_json::json!(request_id),
-            jsonrpc: Some("2.0"),
-        },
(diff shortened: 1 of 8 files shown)


glados-core/src/jsonrpc.rs
```diff
@@ -120,109 +153,246 @@ pub struct Content {
...
-            return Err(JsonRpcError::Empty)?
-        };
-        Ok(obj?)
+impl PortalResponse {
+    /// Creates a Portal response received via HTTP.
+    fn from_http_str(received: &str) -> Result<Self, JsonRpcError> {
+        if received.eq("0x") {
+            return Ok(PortalResponse::ContentAbsent);
+        }
+        let json = serde_json::from_str(received).map_err(|e| JsonRpcError::InvalidJson {
```
kdeme, line 171: I believe there is a mistake here. I've tested this and it always fails with `InvalidJson` error.
It appears that the `received` string is already the value of the result field of the json rpc response object.
So this parsing step can/should be dropped? That is, I could make it work in a quick hacky version with `NonNullHttp(String)` in the `PortalResponse` enum.

perama-v: Thanks for testing. I'll look into it

perama-v: Ok I have changed it so that the HTTP response is now parsed directly into a `serde_json::Value` type to resolve this.

kdeme (approved):
I've been running this PR in combination with Fluffy for some hours without issues.

PR #94, opened 2023-03-22
Add support for panda ops provider

njgheorghita:
In order to add support for glados in our devops pipeline, it makes sense for it to be able to be configured to use the Pandaops nodes as a web3 data provider.

Tested this locally, and confirmed that it's able to retrieve data from pandaops.


njgheorghita:
ping @perama-v @pipermerriam for :eyes:

perama-v (approved):

mrferris (approved):
👍

I don't recall having seen a `Cargo.lock` diff that big with so few `Cargo.toml` diffs. FWIW, I checked and build time appears unaffected.

Issue #93, opened 2023-03-21
Number of audit tasks generated-per-min is not configurable


perama-v:
### Description

`glados-audit` can be viewed as a funnel as follows:

```mermaid
flowchart TD
subgraph generate[Trigger every AUDIT_SELECTION_PERIOD_SECONDS]
s1[Strategy Latest]
s2[Strategy Random]
s3[Strategy Random]
end
s1  & s2 & s3 --> |send KEYS_PER_PERIOD tasks|chan[Audit task channel]


chan --> |take 1 task|a1 & a2 & a3 & a4
subgraph fulfill[Continuously replenish with new threads once they complete]
a1[Auditing thread 1]
a2[Auditing thread 2]
a3[...]
a4[Auditing thread CONCURRENCY]
end
a1 & a2 & a3 & a4 --> node[Portal node]
```

At present, the CLI can control the throughput as follows:
- `--concurrency <n>` flag controls the maximum funnel output rate.
- `--strategy <strat>` flag controls the nature of tasks generated (limited effect on throughput. E.g., setting multiple `--strategy random`)

The two variables that control the maximum funnel input rate are:
- `KEYS_PER_PERIOD`. Currently hard coded as 10.
- `AUDIT_SELECTION_PERIOD_SECONDS`. Currently hard coded as 120 (seconds)

Thus max audits per minute can be calculated:
- one active strategy, the funnel is filled at `10/120 * 60 = 5` tasks (individual content key audits) per minute.
- Currently the default is three active strategies and the funnel is filled at `3 * 10/120 * 60 = 15` tasks (individual content key audits) per minute.

Noting that observed audits/min rate will be lower because audits that timeout against a portal node are not recorded as pass/fail.

The funnel has a "rim height" set to overflow at 100 pending tasks. That is, when the channel has 100 pending tasks, new
tasks generated at this point will be discarded.
 
### Resolution
Expose funnel input control from the CLI. Options flag to:
1. Expose `KEYS_PER_PERIOD` variable.
2. Expose `AUDIT_SELECTION_PERIOD_SECONDS` variable. 
3. Expose `KEYS_PER_PERIOD` and ` AUDIT_SELECTION_PERIOD_SECONDS` variables.
4. New `--max-task-rate <n = max audits per min>` flag that controls maximum audits per minute that are generated. 
    a. Titrate `AUDIT_SELECTION_PERIOD_SECONDS` to `n`, taking into account number of strategies and `KEYS_PER_PERIOD`
[...]

Issue #88, opened 2023-03-20
Support for running multiple clients for glados-audit


pipermerriam:
With #85 being added, it seems we now will want the ability to run `glados-audit` with multiple clients.

- Change the audit schema to allow storing information about the client that was used for the audit.  I suggest we store both a reference to the `ENR` record and the client version string.
- Change to allow "multiple" clients (requiring at least 1) to be specified via `--ipc-path` or `--http-url`
- Change the audit process to round-robin audits across all of the available clients

pipermerriam:
Ok, I think I know how I want to do this.

first, I'd like to implement the ability to do

- `--portal-client http://127.0.0.1:8545`
- `--portal-client ipc:////path/to/portal_jsonrpc.ipc`

And on the rust side, it would be nice to be able to do something like:

```
HistoryClient::from(portal_client_url)
```

Then we need a small abstraction of some kind to allow us to "pool" multiple client such that we can use them like a connection pool.

KolbyML:
https://github.com/ethereum/glados/pull/110 closes this

PR #95, opened 2023-03-23
add strategy weights

perama-v:
### Issue addressed

- #93


perama-v:
> Help me understand why we are adding individual concurrency controls for each strategy rather than a single global concurrency level. I'm not seeing the need or benefit.

The concurrency controls the rate that audits are performed - by having concurrency control for each strategy one can control the rate for each strategy. 

In trying to address issue # 93, the solution seemed most simple (one funnel per strategy, one outflow rate control per funnel, keep each funnel full if possible). For example: `oldest` at 5 concurrent requests, `random` at 50 concurrent requests.

If instead each strategy has a weight (any integer, uncapped) that is relative to other strategies then there is control on both ends of the funnel (variable inflow rate and variable outflow rate, making it hard to predict the real rate). If weights were capped to sum to 100 there was no control over the maximum throughput rate without an additional flag. 

Maybe I am missing what you had in mind as the end goal, happy to rework it

pipermerriam:
The main opposition that I have to individual concurrency levels for each strategy is that I want a system that allows for our global concurrency limit to be consistently saturated.  For example, if we have audited all of the `latest` stuff, I would want the additional concurrency capacity to be usable by `random` or `unaudited`.  This is the main reason why I like a architecture of a single global `--audit-concurrency` flag and having many streams all feeding into the same worker pool.

I'm currently :-1: on having each stream of audits be consumed by a separate worker pool.

As for delivering the desired functionality, I believe this is my end goal and I can be happy with most solutions that get us here.

1. `glados-audit` is able to generally utilize the full concurrency of the connected node.
2. strategies like `latest` which produce jobs on a relatively fixed rate can be reliably allocated priority so that they are consistently fully processed.
3. strategies like `random` that are effectively *infinite* can utilize the excess concurrency when other higher priority jobs have been fully processed.

I have ideas on how this would be implemented (that I'll lay out below), but feel free to implement anything that satisfies these requirements.

My "imagined" architecture looks roughly like this.

- Each strategy gets it's own `tokio::mpsc::channel` with a relatively small buffer size.  Each strategy fills its channel at the maximum rate that it can.
- Each strategy can be configured with a `--strategy-XXX-weight=N` which dictates its relative weight to the other strategies.
- We have a long running "collation" process that consumes jobs from the individual channels at their relative weights (see below) and feeds them into a single global `tokio::mpsc::channel`
- Another long running process consumes jobs from this global channel and feeds them into the threaded worker processes.

The *magic* occurs in the collation process.

In the "normal" case, all of the individual strategy channels have items in them available to be pulled. We'd simply pull from each `mpsc::Receiver` proportional to the configured weights for each strategy, with the rate limiting happening when the buffer size of the global audit channel is full.

In the non-normative case, one of the individual channels will be empty, which would happen in cases like `latest` having already produced jobs for all known new content.  In this case, that channel would be skipped once it was determined it was empty, and the collation process would just move onto pulling jobs from the other channels using the same proportion rules.

I'm not sure if this all makes much sense.  If the total amount of functionality here is bigger than is appropriate for one PR, I'm fine with this getting carved into smaller pieces.  Maybe also worth noting #88 is going to be a priority soon-ish and it might end up interacting with this code as well.

perama-v:
> My "imagined" architecture looks roughly like this.

Ok, here is my understanding of the structure:
```mermaid
---
title: Concurrency weights (latest=1, random=2, oldest=N)
---
flowchart TD
subgraph generate[Strategies fill up their channels as appropriate]
s1[Strategy Latest]
s2[Strategy Random]
s3[Strategy Oldest]
end
s1 --> |send tasks to fill up channel|chan1[Audit task channel 1]
s2 --> |send tasks to fill up channel|chan2[Audit task channel 2]
s3 --> |send tasks to fill up channel|chan3[Audit task channel 3]

chan1 --> |take 1 task|coll
chan2 --> |take 1 task|coll & coll
chan3 --> |take 1 task|coll & coll & coll & coll
chan3 --> |... n|coll
coll[Collate proportional to weights]

coll --> q1 & q2 & q3 & q4 & q5
q1[query 1]
q2[query 2]
q3[query 3]
q4[query 4]
q5[...]
subgraph query[Parallel queries]
q1 & q2 & q3 & q4 & q5 --> node[Portal node]
end
```
With `random=2` and `latest=1`, The collator takes from the `random` channel twice as often as `latest`

glados-audit/src/lib.rs
```diff
@@ -122,19 +151,50 @@ pub async fn run_glados_audit(conn: DatabaseConnection, config: AuditConfig) {
...
+            let quota = tasks.weight * cap / total_weight;
+            debug!(strategy=?tasks.strategy, quota=quota, "collating strategies");
+            for _ in 0..quota {
+                let Some(task) = tasks.channel_recv.recv().await else {continue};
+                if let Err(err) = collation_tx.send(task).await {
+                    error!(err=?err, strategy=?tasks.strategy, "could not move task for collation")
+                }
+            }
+        }
+        sleep(Duration::from_millis(5000)).await;
```
pipermerriam, line 180: This seems like it would be a bit more ergonomic to use an interval here so that we can define it once and just call `interval.tick()`.

pipermerriam:
looks good. Will experiment with getting this merged and deployed Monday.

glados-audit/src/lib.rs
perama-v: Good idea

PR #98, opened 2023-04-03
Add `glados-cartographer`

### LINKED ISSUE #43: Network explorer
Need to write the network explorer pieces.

- process that regularly (every N minutes) "walks" the network using RFN (recursive find nodes) to enumerate all knowable ENR records. 
- a process that looks up ENR records from the database and uses PING messages to determine "liveliness".  At present, we will fail on any node behind a NAT since we don't have traversal.

Some web views that allow exploration of this data.

pipermerriam:
fixes #43 

### What is wrong

Glados needs to monitor the DHT as well

### How was it fixed

This will start by implementing the most basic version of this, which is a long running loop that picks a random `node-id` and then does a `portal_historyRecursiveFindNodes` call, takes all of the result ENR records, and then sends them a PING, recording database records for all nodes that respond with a PONG.

(The PING/PONG may be un-necessary since these nodes *should* have already had a liviliness check performed upon them during the RFN)


pipermerriam:
This PR made me aware that we are still using our "local" implementation of the portal client and that we should switch to the one exposed by `ethportal-api`

pipermerriam:
blocked by https://github.com/ethereum/trin/issues/628

pipermerriam:
ok, basic version working.  Need to remove `dht_audit` model, and fix the migrations that I mucked with.  This iteration simply "probes" parts of the network to build up a view of seen/known ENR records.

pipermerriam:
@njgheorghita this is really fugly, but... poking you for review to get you looking at glados stuff.

PR #99, opened 2023-04-05
Validate received content

perama-v:
### Issue addressed

- Progress towards #96 

### Description

When performing an audit, if content is received, attempt to 
- Check the structure of the content is correct (encoding)
- The second part of the check is addressed in 
    - #100 

   
### Other changes

Cargo.lock: Updated to latest ethportal-api via `cargo update -p ethportal-api` to get HistoryContentValue. Corresponding
changes to import paths for existing ethportal-api.

PR #100, opened 2023-04-05
add correctness check to audited content for header-related content

### LINKED ISSUE #96: Local validation of content
Currently glados is happy with a non-zero response when it audits content.

This should be changed to actually check that the content returned from the network correctly passes validation.

- For headers, we should re-construct the RLP header and verify it hashes to the block hash.
- For bodies we should reconstruct the transaction and uncle tries and verfiy them against the corresponding headers fields
- For receipts we should reconstruct the trie and verify it against the corresponding header field
- For the accumulator, we should verify that the epoch hash matches the one from the master accumulator

perama-v:
### Issue addressed

- Progress towards #96
 
### Dependencies

- Requires #99

### Description

For content that has been received from a Portal node as part of a single content audit, a check is performed. 

The check is the second part of the check (part 1 was structural check). This part checks that the content value is correct with respect to the content key it is connected to. 

A new CLI flag and configuration is added to allow specification of a trusted provider to `glados-audit`

If required, the provider is called to retrieve data necessary for validation. For example, getting the true receipts hash
to confirm that the hash of the receipts retrieved from the Portal node is correct. 

### Epoch accumulator omission

The epoch accumulator correctness check is omitted to limit the scope of the PR. It checks the structure and then issues a warning about an unimplemented correctness check. 

The implementation may be the subject of a different PR.

### Checklist

- [x] CLI
- [x] Local custom HTTP provider implementation
- [x] Pandaops config implementation
- [x] Correctness checks
    - [x] Header 
    - [x] Body
    - [x] Receipts
    - ~~Epoch accumulator~~ - emits unimplemented warning.
- [x] Compact warnings/messages using an error enum 
- [ ] Testing - retrigger CI testing after merge of PR 99
- [x] Tidy commits


perama-v:
Ready for review, changes are isolated from dependency PR 99 for easier review. Left as draft as CI will fail until 99 is merged

pipermerriam:
oops, I have to do a hard rollback here.... stay tuned

PR #101, opened 2023-04-05
Piper/add web explorer view for network node and enrs

pipermerriam:
based on #98


pipermerriam:
@njgheorghita this is ugly but demonstrative of how the web-views are done/rendered.

PR #102, opened 2023-04-07
96 add correctness check

pipermerriam:
This is me re-opening https://github.com/ethereum/glados/pull/100

@perama-v feel free to re-open your own version of this.


perama-v:
Would someone mind also giving this a run and checking that validation is working as expected?

I have also included pandaops connectivity with https://github.com/ethereum/glados/pull/102, but have not validated it works (I just copied from pandops setup elsewhere). Ideally this PR should be tested against that to make sure that it works before updating production glados.

PR #109, opened 2023-04-19
Implement full DHT census

pipermerriam:
### What is wrong

We need `glados-cartographer` to perform full census walking of the network.

### How was it fixed.

The `glados-cartographer` process now performs a full census of the network every 15 minutes.  We still aren't saving them to the database (for example we should be recording the data_radius)...  but this is a start.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1432,7 +1432,7 @@ dependencies = [
  "clap 4.2.1",
  "enr 0.7.0",
  "ethereum-types 0.14.1",
- "ethportal-api",
+ "ethportal-api 0.2.0",
  "glados-core",
  "migration",
  "sea-orm",
@@ -1574,7 +1574,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "219675dc3593f2b99d5cc22d8bbaf956059a66b7eb69444d8502e2b9f5199814"
 dependencies = [
  "hashbrown 0.3.1",
- "keccak-hash",
+ "keccak-hash 0.8.0",
  "log",
  "parking_lot 0.8.0",
  "rlp 0.3.0",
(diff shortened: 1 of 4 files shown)


pipermerriam:
This implements the basic functionality, blocked by having a basic system for recording client version string info, and then having a db model for saving census info.

pipermerriam:
I've got an issue that I'm really not sure how to correctly fix.

```
error[E0597]: `limiter` does not live long enough
   --> glados-cartographer/src/lib.rs:289:22
    |
289 |         let permit = limiter.acquire().await.expect("Error acquiring lock");
    |                      ^^^^^^^^^^^^^^^^^ borrowed value does not live long enough
290 |         tokio::spawn(do_liveliness_check(enr, tx.clone(), census.clone(), config.clone(), conn.clone(), permit));
    |                      ------------------------------------------------------------------------------------------ argument requires that `limiter` is borrowed for `'static`
291 |     }
292 | }
    | - `limiter` dropped here while still borrowed

error[E0597]: `limiter` does not live long enough
   --> glados-cartographer/src/lib.rs:349:22
    |
349 |         let permit = limiter.acquire().await.expect("Error acquiring lock");
    |                      ^^^^^^^^^^^^^^^^^ borrowed value does not live long enough
350 |         tokio::spawn(do_routing_table_enumeration(enr, tx.clone(), census.clone(), config.clone(), permit));
    |                      ------------------------------------------------------------------------------------- argument requires that `limiter` is borrowed for `'static`
351 |     }
352 | }
    | - `limiter` dropped here while still borrowed

For more information about this error, try `rustc --explain E0597`.
```

I understand the error, but I don't understand how to fix it.  It seems I need to maybe do something with the lifetime of the `limiter` object, but that's about as far as I've been able to figure out.

perama-v:
Could you push the changes where this error is occurring?

pipermerriam:
I ended up fixing it this morning.  Bad commit hygene on my side and don't really have any easy way to expose the history correctly without dumb git futzing....

pipermerriam:
replaced by #117

PR #111, opened 2023-04-24
Cleanup strategy weights code

pipermerriam:
### What was wrong

There were a few things in the code for strategy weights that I wanted to cleanup

### How was it fixed

I did some minor refactoring.


perama-v:
I like it, much cleaner 

Added a small change to differentiate two error variants:
- https://github.com/pipermerriam/glados/pull/1

perama-v:
> Added a small change to differentiate two error variants:

This turned out to be a bad idea because it is necessary to ignore unimplemented strategies.

perama-v (approved):

PR #113, opened 2023-04-27
use git for ethportal-api crate

perama-v:
## What was wrong

`ethportal-api` was behind that available on `trin`. Keeping up to date requires action on the `trin` side, which has included friction when `trin` publishing is not possible without changes to the `ethportal-api` crate.

## What was changed

Move ethportal-api to point to git instead of crates so that glados can use the latest ethportal-api by running:
```command
cargo update -p ethportal-api
```


mrferris:
Done in #241

PR #114, opened 2023-05-01
Add ability to query using xor distance metric for nodes.

pipermerriam:
### What was wrong

Glados needs the ability to navigate the relationships between nodes and content using the distance netric

### How was it fixed

This adds the ability to query nodes as they relate to each other through the xor distance metric.  This is done by storing the high 63 bits as an i64 which is the largest data types.  This allows us to have the database do XOR ordering queries.


pipermerriam:
Well this is pretty cool.  Once I have this merged we'll be able to list the nearest known content and other nodes when viewing either content or a node.

glados-web/src/lib.rs
```diff
@@ -27,6 +30,25 @@ pub async fn run_glados_web(config: Arc<State>) -> Result<()> {
...
+    for node_model in nodes_with_zero_high_bits {
+        let raw_node_id = U256::from_big_endian(node_model.get_node_id().raw().as_slice());
+        let node_id_high: i64 = (raw_node_id >> 193).as_u64().try_into().unwrap();
+
+        let mut node: entity::node::ActiveModel = node_model.into();
+        let previous_value = node.node_id_high;
+        node.node_id_high = Set(node_id_high);
+        let updated = node.update(&config.database_connection).await?;
+        info!(row.id=?updated.id, old=?previous_value, new=?updated.node_id_high, "Setting high bits");
+    }
```
pipermerriam, line 50: This section here is effectively a data migration that just needs to run once in our production environment.  I looked at making a separate command to do this but it involved a lot of overhead.  My plan is to leave this in place until I can verify it's been run in production, after which I'll PR to remove it.

entity/Cargo.toml
```diff
@@ -12,11 +12,15 @@ authors = ["Piper Merriam <piper@pipermerriam.com>"]
 anyhow = "1.0.68"
 chrono = "0.4.22"
 clap = { version = "4.0.24", features = ["derive"] }
-enr = "0.7.0"
+enr = "0.8.1"
 ethereum-types = "0.14.0"
-ethportal-api = { git = "https://github.com/ethereum/trin" }
+ethportal-api = "0.1.6"
```
ogenev, line 17: What motivates switching from git path (latest `ethportal-api` unpublished version)  to the latest cargo version?

pipermerriam: I was minorly bothered by the git version being "imprecise" since it doesn't anchor to a specific piece of code, and it seemed like I was able to get everything working in an acceptable way with the released versions.

entity/src/node.rs
```diff
@@ -33,6 +56,48 @@ impl Related<super::record::Entity> for Entity {
...
+    let node_id_high: i64 = (raw_node_id >> 193).as_u64().try_into().unwrap();
+
+    let nodes = Entity::find()
+        .column_as(
+            Expr::cust_with_values(
+                // This line has to be changed depending on whether we are using sqlite or
+                // postgres.
+                // sqlite:   "(\"node\".\"node_id_high\" | ?1) - (\"node\".\"node_id_high\" & ?2)",
+                // postgres: "(\"node\".\"node_id_high\" | $1) - (\"node\".\"node_id_high\" & $2)",
+                "(\"node\".\"node_id_high\" | $1) - (\"node\".\"node_id_high\" & $2)",
```
mrferris, line 88: You could generate this with:

`let expression = if conn.get_database_backend() == DatabaseBackend::Sqlite { ... } else { ... }`

PR #119, opened 2023-05-11
Reset SQL migrations

mrferris:
Creates all tables with their needed columns, rather than adding columns later on. Will require a DB wipe.


migration/src/m20230511_104804_create_node.rs
```diff
@@ -7,36 +7,40 @@ pub struct Migration;
 impl MigrationTrait for Migration {
     async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
         manager
-            .alter_table(
-                Table::alter()
+            .create_table(
+                Table::create()
                     .table(Node::Table)
-                    .add_column_if_not_exists(
+                    .if_not_exists()
```
pipermerriam, line 13: I'm wondering if this is the right thing?  I think we'd want this to throw an error if the table already existed.

migration/src/m20230511_104814_create_content.rs
```diff
@@ -0,0 +1,70 @@
...
+                            .timestamp_with_time_zone() // chrono::DateTime<FixedOffset>
+                            .not_null(),
+                    )
+                    .index(
+                        Index::create()
+                            .unique()
+                            .name("idx-unique-id-protocol-and-key") // Triple column constraint
+                            .col(Content::ProtocolId) // 1
+                            .col(Content::ContentKey) // 2
+                            .col(Content::ContentId), // 3
```
pipermerriam, line 47: I "think" we need another uniqueness index that only covers `(protocol-id, content-key)`.  

One of the plans that I have for the content key system in the future is to have multiple content keys that map to the same content-id.  The use case is having a content key for an individual transaction, that still maps to the same content-id as the block body, so that the nodes holding the block body can return only a single transaction from the body rather than the whole body payload.

migration/src/m20230511_104814_create_content.rs
```diff
@@ -0,0 +1,70 @@
...
+                            .integer() // i32
+                            .not_null(),
+                    )
+                    .col(
+                        ColumnDef::new(Content::ContentKey)
+                            .binary_len(33) // 33 bytes
+                            .not_null(),
+                    )
+                    .col(
+                        ColumnDef::new(Content::ContentId)
```
pipermerriam, line 32: I think we need an index on `(protocol-id, content-id)` as this will be a common query for looking up content.

migration/src/m20230511_104814_create_content.rs
```diff
@@ -0,0 +1,70 @@
...
+                            .binary_len(32) // 32 bytes
+                            .not_null(),
+                    )
+                    .col(
+                        ColumnDef::new(Content::FirstAvailableAt)
```
pipermerriam, line 37: I think we need an index on `(first_available_at, procotol-id)` which should be sufficient for both bare filtering on `first_available_at` **and** on queries that involve both fields.

migration/src/m20230511_104814_create_content.rs
```diff
@@ -0,0 +1,70 @@
...
+                            .primary_key(),
+                    )
+                    .col(
+                        ColumnDef::new(Content::ProtocolId)
+                            .integer() // i32
+                            .not_null(),
+                    )
+                    .col(
+                        ColumnDef::new(Content::ContentKey)
+                            .binary_len(33) // 33 bytes
```
pipermerriam, line 28: This is wrong.  This field should be an unbounded binary blob.  It' is coincidence that all of our current content keys are the same length.

migration/src/m20230511_104814_create_content.rs
```diff
@@ -0,0 +1,70 @@
...
+                    .col(
+                        ColumnDef::new(Content::Id)
+                            .integer() // i32
+                            .not_null()
+                            .auto_increment()
+                            .primary_key(),
+                    )
+                    .col(
+                        ColumnDef::new(Content::ProtocolId)
+                            .integer() // i32
```
pipermerriam, line 23: This *might* be more correct as an unsigned...  unclear how much this matters.

migration/src/m20230511_104830_create_content_audit.rs
```diff
@@ -0,0 +1,97 @@
...
+                        ForeignKey::create()
+                            .name("FK_contentaudit_content_key")
+                            .from(ContentAudit::Table, ContentAudit::ContentKey)
+                            .to(Content::Table, Content::Id)
+                            .on_delete(ForeignKeyAction::Cascade)
+                            .on_update(ForeignKeyAction::Cascade),
+                    )
+                    .col(
+                        ColumnDef::new(ContentAudit::CreatedAt)
+                            .timestamp_with_time_zone() // chrono::DateTime<FixedOffset>
```
pipermerriam, line 36: I think this is a copy/paste artifact and that this field should be a `DateTime<Utc>`

migration/src/m20230511_104830_create_content_audit.rs
```diff
@@ -0,0 +1,97 @@
...
+                            .not_null(),
+                    )
+                    .col(ColumnDef::new(ContentAudit::Result).integer().not_null())
```
pipermerriam, line 39: I believe we can do one index `(created_at, result)` which IIUC will support both querying solely based on the `created_at` as well as filtering on both fields.

migration/src/m20230511_104838_create_execution_metadata.rs
```diff
@@ -0,0 +1,72 @@
...
+                            .auto_increment()
+                            .primary_key(),
+                    )
+                    .col(
+                        ColumnDef::new(ExecutionMetadata::Content)
+                            .integer()
+                            .not_null(),
+                    )
+                    .col(
+                        ColumnDef::new(ExecutionMetadata::BlockNumber)
```
pipermerriam, line 27: I suspect we'll want to query on this field, thus it needs a database index.

migration/src/m20230511_104937_create_key_value.rs
```diff
@@ -0,0 +1,57 @@
...
+                    .if_not_exists()
+                    .col(
+                        ColumnDef::new(KeyValue::Id)
+                            .integer()
+                            .auto_increment()
+                            .not_null()
+                            .primary_key(),
+                    )
+                    .col(ColumnDef::new(KeyValue::RecordId).integer().not_null())
+                    .col(ColumnDef::new(KeyValue::Key).binary().not_null())
```
pipermerriam, line 22: We need two indexes here:

- an index on `(record_id, key)` unique index. which enforces the constraint and allows quick querying of all the keys for a given `record_id`.
- an index on only `key` so that we can lookup all the records with a given key.

pipermerriam (changes requested):
so sorry if you already had plans to add these.

mrferris:
> so sorry if you already had plans to add these.

Hadn't thought about most of these indices/queries, so thanks!

Currently seeing sea orm generating invalid sql syntax when creating non-unique indices (which is why they're commented out here):

```

CREATE TABLE "content" (
      "id" serial NOT NULL PRIMARY KEY,
      "protocol_id" integer NOT NULL,
      "content_key" bytea NOT NULL,
      "content_id" bytea NOT NULL,
      "first_available_at" timestamp with time zone NOT NULL,
      CONSTRAINT "idx-unique-id-protocol-and-key" UNIQUE ("protocol_id", "content_key", "content_id"),
      CONSTRAINT "idx-unique-protocol-and-key" UNIQUE ("protocol_id", "content_key"),
      CONSTRAINT "idx-unique-protocol-and-id" UNIQUE ("protocol_id", "content_id"),
      CONSTRAINT "idx-available-and-protocol" ("first_available_at", "protocol_id")
    )

Error: Execution Error: error returned from database: syntax error at or near "("

```

pipermerriam:
I think I ran into this same issue, any chanced you looked upstream to see if this is a known sea-orm issue?

Lets move forward at least with the unique ones and open an issue to figure out the non-unique ones at a later date.

PR #121, opened 2023-05-12
Update SETUP_GUIDE.md to use latest cli options

kdeme:
Setup guide was a little outdated.

### DIFF
--- a/SETUP_GUIDE.md
+++ b/SETUP_GUIDE.md
@@ -15,9 +15,8 @@ This example uses the following:
 
 All commands are issued on `machine-a` unless otherwise stated.
 
-Important notes:
+Important note:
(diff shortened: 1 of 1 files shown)


SETUP_GUIDE.md
```diff
@@ -49,7 +48,7 @@ Start `glados-audit`, which takes monitoring data from the glados database,
 checks if `trin` has it, then records the outcome in the glados database.
 ```command
 ~/glados$ RUST_LOG=debug cargo run -p glados-audit -- \
-    --ipc-path /path/to/trin-jsonrpc.ipc \
+    --portal-client /path/to/trin-jsonrpc.ipc \
```
mrferris, line 51:
```suggestion
    --portal-client ipc:////path/to/trin-jsonrpc.ipc \
```

mrferris (approved):
I should've updated that in #110, good catch. Only thing is that the argument needs the `ipc:///` (or `http://`) prefix, then good to merge 👍

PR #122, opened 2023-05-12
Use client info within dashboard

### LINKED ISSUE #124: Show clients in glados content dashboard

mrferris:
Fixes #124 

- [x] add client info for each audit to dashboard
- [x] refactor `content_dashboard` route function to use [seaorm's DataLoader api](https://www.sea-ql.org/SeaORM/docs/relation/data-loader/), which is 1. more ergonomic for loading related data and 2. is more efficient as it doesn't load duplicate data.  Though I wonder if we should just be doing a big join and load it all in one query rather than 2 or 3 separate calls.

### DIFF
--- a/glados-core/src/jsonrpc.rs
+++ b/glados-core/src/jsonrpc.rs
@@ -223,7 +223,8 @@ impl PortalClient {
             client_url: portal_client_url.clone(),
         };
 
-        let client_info = &api.get_client_version().await?;
+        let client_info = api.get_client_version().await?;
+        let stripped_client_info = strip_quotes(client_info);
 
         let node_info = &api.get_node_info().await?;
 
@@ -235,7 +236,7 @@ impl PortalClient {
 
         Ok(PortalClient {
             api,
-            client_info: client_info.to_string(),
+            client_info: stripped_client_info.to_string(),
             enr,
         })
     }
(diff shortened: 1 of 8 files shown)


glados-core/src/jsonrpc.rs
```diff
@@ -223,7 +223,15 @@ impl PortalClient {
             client_url: portal_client_url.clone(),
         };
 
-        let client_info = &api.get_client_version().await?;
+        let mut client_info = api.get_client_version().await?;
+        // Strip quotes from client info result if needed
+        if client_info.starts_with('"') && client_info.ends_with('"') {
```
pipermerriam, line 228: This looks like a nice candidate for a small function that could have tests?

glados-web/src/routes.rs
```diff
@@ -216,123 +216,206 @@ pub async fn content_dashboard(
...
+    num_audits: u64,
+    conn: &DatabaseConnection,
+) -> Result<Vec<AuditTuple>, StatusCode> {
+    let recent_audits: Vec<content_audit::Model> = content_audit::Entity::find()
+        .order_by_desc(content_audit::Column::CreatedAt)
+        .filter(content_audit::Column::Result.eq(AuditResult::Success))
+        .limit(num_audits)
+        .all(conn)
+        .await
+        .map_err(|e| {
```
pipermerriam, line 317: Any opinions or thoughts on `map_err` vs `match` against `Ok/Err` values?  Is there an idiomatic correct approach here in rust?  I think this was introduced by @perama-v .  Independent on my own I started using `match` statements more often.  I'm not sure I have any strong opinions yet for one-or-the-other.

perama-v: Agree. I like .map_err() to reduce nesting that match causes. It's nice when you don't want to modify the x in the Ok(x) variant, but you want to modify the e in the Err(e) variant.

PR #125, opened 2023-05-16
Add specific single key audit

njgheorghita:
Add a subcommand to glados that executes a single one-off audit for previously audited content keys. This command is only functional for on-disk databases, passed in via a cli arg, and will not be effective on in-memory dbs. 

- Skipped support for all valid content keys (not just previously audited ones) to avoid complexity involved w/ looking up the block number via `web3` & inserting it into db. I figured if the use case is important, it can be added later
- Spent some time playing around with various configurations of the cli parser / subcommand structure. In the end this was the simplest design I could get to work.

### DIFF
--- a/entity/src/content_audit.rs
+++ b/entity/src/content_audit.rs
@@ -34,6 +34,8 @@ pub enum SelectionStrategy {
     /// 1. Not yet audited.
     /// 2. Sorted by date entered into glados database (oldest first).
     SelectOldestUnaudited = 3,
+    /// Perform a single audit for a previously audited content key.
+    SpecificContentKey = 4,
 }
 
 impl AuditResult {
@@ -156,6 +158,7 @@ impl SelectionStrategy {
             SelectionStrategy::Random => "Random".to_string(),
             SelectionStrategy::Failed => "Failed".to_string(),
             SelectionStrategy::SelectOldestUnaudited => "Select Oldest Unaudited".to_string(),
+            SelectionStrategy::SpecificContentKey => "Specific Content Key".to_string(),
         }
     }
 }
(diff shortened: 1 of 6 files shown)


entity/src/content_audit.rs
```diff
@@ -34,6 +34,8 @@ pub enum SelectionStrategy {
     /// 1. Not yet audited.
     /// 2. Sorted by date entered into glados database (oldest first).
     SelectOldestUnaudited = 3,
+    /// Perform a single audit for a previously audited content key.
+    SpecificContentKey = 4,
```
pipermerriam, line 38: Maybe `Manual`

glados-audit/src/cli.rs
```diff
@@ -42,6 +42,18 @@ pub struct Args {
...
+    #[command(subcommand)]
+    pub subcommand: Option<Command>,
+}
+
+#[derive(Subcommand, Debug, Eq, PartialEq, Clone)]
+pub enum Command {
+    /// Run a single audit for a specific, previously audited content key.
+    Audit {
+        content_key: String,
+        portal_client: String,
```
pipermerriam, line 54: Can these not live in the parent so that they are only defined once?


```
glados-audit --portal-client http://arst --database-url sqlite://arst audit --content-key
```

glados-audit/src/cli.rs
```diff
@@ -42,6 +42,18 @@ pub struct Args {
...
     pub portal_client: Vec<String>,
+    #[command(subcommand)]
+    pub subcommand: Option<Command>,
+}
+
+#[derive(Subcommand, Debug, Eq, PartialEq, Clone)]
+pub enum Command {
+    /// Run a single audit for a specific, previously audited content key.
+    Audit {
+        content_key: String,
```
pipermerriam, line 53: Can this be done in such a way that we can provide the argument multiple times to allow us to run a batch of audits?

pipermerriam (approved):
Looks great, do as you will with the feedback I left, none of it is required.

glados-audit/src/cli.rs
mrferris: I think the issue is then that it's ambiguous which client you'd like to use for the audit if there's multiple

mrferris (approved):

PR #134, opened 2023-06-09
Improve scalability of stat calculations

mrferris:
When calculating the number of content/audits in the past hour/week/day, we were loading every single row and then counting the size of the returned vec, rather than just having the DB count it for us via `count`. 

Previously our stat query times looked like this on prod (with 3.9m audit rows):

Hour stats: Content: 134 ms, Audits: 542 ms, Passes: 153 ms
Day stats:   Content: 142 ms, Audits: 1844 ms, Passes: 240 ms
Week stats:Content: 358 ms, Audits: 2671 ms, Passes: 452 ms

with this change they look like:

Hour stats: Content: 189 ms, Audits: 364 ms, Passes: 366 ms
Day stats:   Content: 182 ms, Audits: 355 ms, Passes: 361 ms
Week stats:Content: 183 ms, Audits: 362 ms, Passes: 421 ms

### DIFF
--- a/glados-web/src/routes.rs
+++ b/glados-web/src/routes.rs
@@ -571,34 +571,32 @@ async fn get_audit_stats(period: Period, conn: &DatabaseConnection) -> Result<St
     let cutoff = period.cutoff_time();
     let new_content = content::Entity::find()
         .filter(content::Column::FirstAvailableAt.gt(cutoff))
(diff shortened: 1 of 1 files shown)


mrferris:
Merging, as this has been running in prod successfully

PR #138, opened 2023-06-28
`ethportal_api` Integration

mrferris:
Currently glados uses a custom json rpc client and deserializes the responses to custom types.

This PR integrates a `jsonrpsee` client and replaces custom types with `ethportal_api` types. 

WIP

Implements #123

PR #139, opened 2023-06-28
Expect discv5::enr::NodeId in NodeInfo repsonse

mrferris:
Due to a change in `ethportal_api`, trin's  representation of a NodeId in a `NodeInfo` response changed.

This happening in the future will be avoided once #138 is merged


mrferris:
On second thought, this violates the portal spec, which for this node ID field reads: 

   _description: Hex encoded `NodeId` of an ENR (a 32 byte identifier).
   pattern: ^0x[0-9a-f]{64}$_

whereas `discv5::enr::NodeId` is serializing to an array of bytes.

PR #142, opened 2023-07-19
Bulk fetch block hashes to allow for auditing old content keys

mrferris:
**What was wrong?**
Glados could not audit any content keys for blocks that existed before `glados-monitor` began running

**How was it fixed?**
Implemented `bulk-download-block-data` subcommand for `glados-monitor` with `beginning`, `end`, and `concurrency` arguments for block number range to fetch.

With `concurrency` set to 100 locally, I am seeing a retrieval/storage rate of 3.8 million blocks per day, so it should take 4.6 days to enter every content key into the glados DB. 

- [x] Batching
- [x] Retries


glados-monitor/src/lib.rs
```diff
@@ -243,3 +253,88 @@ pub async fn import_pre_merge_accumulators(
...
+    end: u64,
+    provider_url: String,
+    concurrency: u32,
+) -> Result<()> {
+    if beginning > end {
+        error!("Beginning block number must be less than or equal to end block number");
+        Err(anyhow!(
+            "Beginning block number must be less than or equal to end block number"
+        ))?;
+    }
```
carver, line 270: *nit* this wording feels awkward. I'd say usually the end number is the one that needs to change. Like the end block number must be greater than or equal to the beginning.

Also, when raising this error, it should will be logged out by the caller. [looking it up...] Oh, the result is being totally ignored! Oof.

Ok, so I'd say drop the `error!` log here and we should modify the `tokio::select` to check for errors (checking both the outer layer for the spawn being cancelled, and the inner layer for the command failing). Then display the error message instead of always showing "Command completed, exiting".

glados-monitor/src/lib.rs
```diff
@@ -243,3 +253,88 @@ pub async fn import_pre_merge_accumulators(
...
+    info!(
+        beginning = beginning,
+        end = end,
+        provider_url = provider_url.as_str(),
+        "Starting bulk download of block data",
+    );
+
+    let w3 = panda_ops_web3(provider_url).expect("Failed to connect to PandaOps");
```
carver, line 278: *nit* How about propagating the error with `?` instead of expect? Then we can use the new top-level error display.

I think it's probably better to show a one-liner (eg~ "PANDAOPS_CLIENT_ID env var not set.") than to panic with the more generic "Failed to connect to PandaOps" and a stack trace.

glados-monitor/src/lib.rs
```diff
@@ -243,3 +253,88 @@ pub async fn import_pre_merge_accumulators(
...
+                    };
+
+                    let block_number =
+                        i32::try_from(block_number).expect("Block num does not fit in i32.");
+                    store_block_keys(block_number, block_hash.as_fixed_bytes(), &conn).await;
+                })
+            })
+            .collect();
+
+        join_all(join_handles).await;
```
carver, line 317: Since this will always wait for the slowest request to complete before starting the next batch of requests, it will end up under-utilizing the provided concurrency (especially because the slowest one is sleeping and retrying).

We don't necessarily need to focus on maximizing throughput here, but it's at least worth mentioning in a comment that it's a known performance limitation.

But... I don't think it would be too much work to do the maximization now. Something like: 
- before the loop, prepare a semaphore, sized by the concurrency
- iterate through all block numbers one by one
- at the beginning of the loop, [acquire an owned semaphore permit](https://docs.rs/tokio/latest/tokio/sync/struct.Semaphore.html#method.acquire_owned) (in order to block when all `concurrency` tasks are busy)
- launch the requesting task, passing in the permit
- at the end of the task, drop the permit

That should make sure we always have `concurrent` tasks running.

carver (approved):
👍🏻 to merge when you feel everything is addressed!

glados-monitor/src/cli.rs
```diff
@@ -36,4 +36,16 @@ pub enum Commands {
         #[arg(short, long)]
         path: PathBuf,
     },
+
+    /// Imports blocks from a remote provider
+    BulkDownloadBlockData {
+        #[arg(short, long)]
+        beginning: u64,
```
pipermerriam, line 43: I would suggest this be `start-block-number` and `end-block-number` so that the arguments are self documenting.

glados-monitor/src/lib.rs
```diff
@@ -86,39 +94,45 @@ async fn retrieve_new_blocks(
...
+    let block = w3
+        .eth()
+        .block(BlockId::from(block_number))
+        .await
+        .map_err(|e| anyhow!("Failed to retrieve block: {}", e))?
+        .ok_or_else(|| anyhow!("No block found at {block_number}"))?;
+
+    let block_hash = block
+        .hash
+        .ok_or_else(|| anyhow!("Fetched block has no hash (skipping)"))?;
```
pipermerriam, line 125: *nitpick* errors like this that don't expose the payload that the error came from tend to be unhelpful in actual production situations.  In the event that we hit this error in a log, we're gonna want to see the debug representation of `block` in order to understand why there was no `hash` field....  maybe this should include a debug printout of the `block` variable?

glados-monitor/src/lib.rs
```diff
@@ -170,7 +180,7 @@ async fn store_content_key<T: OverlayContentKey>(
 /// Helper function for common error pattern to be logged.
 fn log_record_outcome<T: OverlayContentKey>(key: &T, name: &str, outcome: DbOutcome) {
     match outcome {
-        DbOutcome::Success => info!(
+        DbOutcome::Success => debug!(
```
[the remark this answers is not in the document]
pipermerriam: I agree and I'm also kinda-ok with it.  This script is gonna be inherently noisy... unless you wanna do the extra leg-work to do a heartbeat-style log message that aggregates this information.  No strong actual opinions here on which way you take this.

glados-monitor/src/lib.rs
mrferris: @carver Good point. I made that change.

Interestingly, in practice I'm seeing slightly worse performance doing it with a semaphore than I was with `join_all`. 
With a fresh DB (inserting blocks we already have is faster than ones we don't), 10k blocks takes:
- with `join_all`: 3m12s
- with semaphore: 3m42s

The only reason I can think of is that when all of these writes come into the DB with a break after (thanks to `join_all`), the DB implements some efficient batching that can't be done when a constant stream of inserts just keeps coming as with the semaphore approach. 

Curious if you have any thoughts on that.

carver: Hm, yeah ~15% slower is a bit surprising. Though hunting down ~3ms per task may not be the best use of time. Random variation is certainly a possibility. I don't have a reasonable explanation why semaphore would be slower.

In the happy path, where the server is replying quickly and consistently, it makes a lot of sense that they would perform about the same. Using semaphores should really save you when the server is getting slow and inconsistent. Just one out of 100 requests taking 20% longer would overcome this 15% cost, because the other 99 requests would sit idle waiting for the slowest one to complete.

I'm curious to know, how do they both perform at different levels of concurrency, like: 25, 50, 200, 400? From chat, it sounds like the join_all version took the same amount of time at lower concurrency, but semaphore got slower (what was the concurrency on that?).

---

There's a possibility that one of the implementations is not doing what we think it's doing, so a small run with low concurrency and some logs on both variants could be worth building an understanding. Basically just logging individual messages at: (Block # 1 started, Block # 1 retrieved, Block # 1 stored in DB).

I would try with 30 blocks and 3 concurrency in both variants. In the join-all variant, we should never see the 4th block start before 1-3 are complete. In the semaphore variant, the 4th block should start as soon as one of the first 3 completes. It would also be interesting to see the log timestamps, as a cheap measure of performance of each of the steps (how long are the block request and DB insertion steps taking? how much overhead is there: the length between one task ending and the next beginning? etc.).

---

Though I would also be ok just choosing the semaphore approach, accepting the unexplained hit to performance, and hunt it down with profiling later, if you want to move on. I would not be very happy with taking the join-all that works better in the happy case, but we have every reason to think will be much slower in higher-variance situations.

mrferris: Unfortunately it doesn't remain a 15% difference as it scales. 15% was at 10k, at 30k there's a 48% slow down:
```
	join_all:
		real	11m40.236s
		user	4m43.700s
		sys	1m37.186s
	semaphore:
		real	17m21.742s
		user	5m52.260s
		sys	1m57.537s
```

Notably, the semaphore strategy becomes faster than the `join_all` strategy on sqlite. This makes me feel confident that the DB is where this is happening. 

On postgres, the semaphore strategy starts to really crawl to almost one task complete per second at some points while doing 30k. It feels like I'm driving a manual car incorrectly and destroying its transmission. I am theorizing that Postgres likes this pattern of insertions better than the semaphores and I will look for some documentation to support that.

I will lessen the retry interval shorter so that we would need around a 4% failure rate to make the join_all method as slow as the semaphore method. I'm currently seeing a 0% failure rate so that feels comfortable.

carver: Yeah, super interesting that Postgres is slower and sqlite is faster. It might even change as the size of the databases change. There are so many variables involved.


> [This post](https://dba.stackexchange.com/questions/75075/how-do-i-maintain-high-insert-performance-on-postgresql) might provide a clue as to why:

Nice link! Some really interesting things to try there. My intuition leads me more to a lock issue (also mentioned in the top answer), rather than the cache.

After poking around a bit, the other approach I would want to try when optimizing is something that turns the 1-stage pipeline into two stages: load hashes, and insert to DB. The DB insertion might be best with really low concurrency on Postgres, like 2-10, while leaving the web request concurrency at say ~100. We can use a bounded channel to prevent the web requests from racing ahead too far. Also, there's some low-hanging fruit to improve performance by refactoring the query pattern (like we don't really need to do the select before insert).

But... since this is not a heavily recurring activity, I'm okay with just going with the hack you mentioned on the call (implement both strategies, and choose which to use, if connected to postgres or sqlite). But as soon as we start putting more effort into optimizing, then we should either try this 3rd approach mentioned in the paragraph above, and/or stick with the semaphore approach that's slower to start and optimize from there until it beats the join_all. (We both have different intuition about what's slow. The only way we would know for sure is to implement the alternatives, tinker with the configurations, turn on logging, and find out)

> I am theorizing that Postgres likes the join_all pattern of insertions better than the semaphores.

I still refuse to believe that `join_all` will be generally superior to an optimized semaphore approach, without hard evidence. :)  There's nothing special that join_all is doing better than we can do explicitly, while making it also waste less time.  Also, we would need to do same basic logging and testing to make sure both approaches are really doing what we think they are...

But again, I am happy to drop this for now, and pick it back up when we decide that runtime performance of this feature is of higher importance to us. I feel invested enough that I'd be happy to pick up the primary work of making it faster, but I don't want to get sucked into it now, it when it's not a high priority.

So let's 🚢 and move on to top-priority stuff!

mrferris: > I still refuse to believe that join_all will be generally superior to an optimized semaphore approach, without hard evidence. :) There's nothing special that join_all is doing better than we can do explicitly

Yea I didn't mean to imply that `join_all` is a magic incantation, just that the delay that it introduces results in the higher performance of the two approaches considered. I agree that an approach with a dedicated DB worker using a very specific delay/rate/concurrency number found by understanding what's going on inside postgres will perform even better.

And good point about removing the select that's done before the insertion.

PR #144, opened 2023-08-01
Unique index on seq_num prevents new records with same seq-num to be added

KaydenML:
Unique index on seq_num prevents new reocrds with same seq-num to be added


mrferris:
Good catch! You're correct, I'd intended to make that a non-unique index, which I will add back later on.

pipermerriam:
Might be overkill but adding a test for this type of thing will ensure that we don't hit this again.

PR #145, opened 2023-08-01
Styled-Content-dashboard-to-match-theme-of-main-page

KaydenML:
I ended up styling the Content_Dashboard page to match the style of the Main Page that i had styled on my previous PR.

### DIFF
--- a/glados-web/assets/css/content_dashboard.css
+++ b/glados-web/assets/css/content_dashboard.css
@@ -1,14 +1,4 @@
-.dashboard-box {
-    background-color: #ffffff;
-    padding: 10px;
-    border-radius: 10px;
-    box-shadow: 0 7px 25px rgba(0, 0, 0, 0.08);
-    height: 100%;
-    width: 98%;
-}
-
-
 .header {
-    margin-bottom: 10px;
+    margin-bottom: 5px;
     font-family: Arial, sans-serif;
 }
--- a/glados-web/assets/css/content_dashboard.css
+++ b/glados-web/assets/css/content_dashboard.css
@@ -9,6 +9,6 @@
 
 
 .header {
-    margin-bottom: 30px;
+    margin-bottom: 10px;
     font-family: Arial, sans-serif;
 }
--- a/glados-web/assets/css/content_dashboard.css
+++ b/glados-web/assets/css/content_dashboard.css
@@ -1,0 +1,53 @@
+.dashboard-box {
+    background-color: #ffffff;
+    padding: 20px;
+    border-radius: 20px;
+    box-shadow: 0 7px 25px rgba(0, 0, 0, 0.08);
+    height: 275px;
+    width: 1855px;
+}
+.dashboard-box2 {
+    background-color: #ffffff;
+    padding: 20px;
+    border-radius: 20px;
+    box-shadow: 0 7px 25px rgba(0, 0, 0, 0.08);
+    height: 850px;
+    width: 1855px;
+}
+.dashboard-box3 {
+    background-color: #ffffff;
+    padding: 20px;
+    border-radius: 20px;
+    box-shadow: 0 7px 25px rgba(0, 0, 0, 0.08);
+    height: 950px;
+    width: 1855px;
+}
+.dashboard-box4 {
+    background-color: #ffffff;
+    padding: 20px;
+    border-radius: 20px;
+    box-shadow: 0 7px 25px rgba(0, 0, 0, 0.08);
+    height: 950px;
+    width: 1855px;
+}
+.dashboard-box5 {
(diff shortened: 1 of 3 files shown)


glados-web/assets/css/content_dashboard.css
```diff
@@ -0,0 +1,53 @@
+.dashboard-box {
+    background-color: #ffffff;
+    padding: 20px;
+    border-radius: 20px;
+    box-shadow: 0 7px 25px rgba(0, 0, 0, 0.08);
+    height: 275px;
+    width: 1855px;
```
mrferris, line 7: We need to keep everything viewable on a single page, so we don't want to use static widths for our elements.

```suggestion
    width: 100%;
```

glados-web/assets/css/content_dashboard.css
```diff
@@ -0,0 +1,53 @@
...
+}
+.dashboard-box2 {
```
mrferris, line 9: Just have each div be a `.dashboard-box`, we don't need to define a class for every element that will have the same exact features.

mrferris (changes requested):
Having some drop shadow to distinguish different dashboard components is good, though too much empty white space is being added. I would compare the spacing on the left side of and top of the page to the current spacing on glados.ethportal.net and get it close as possible to that so that we aren't regressing in terms of the page's information density.

KaydenML:
image
image

KaydenML:
before image
image

glados-web/templates/content_dashboard.html
```diff
@@ -1,16 +1,21 @@
 {% extends "base.html" %}
 
 {% block title %}Content Dashboard{% endblock %}
-
+{% block head %}
+<link rel="stylesheet" href="/static/css/content_dashboard.css">
```
mrferris, line 5:
```suggestion
    <link rel="stylesheet" href="/static/css/content_dashboard.css">
```

glados-web/assets/css/content_dashboard.css
```diff
@@ -0,0 +1,11 @@
...
+    border-radius: 10px;
+    width: 98% !important;
+    height: 100% !important;
+    border: 2px white;
+}
+
+.header {
+    margin-bottom: 5px;
+    font-family: Arial, sans-serif;
+}
```
mrferris, line 11: Be sure to add a newline at the end of the file to avoid this error/marking

glados-web/templates/content_dashboard.html
```diff
@@ -45,185 +52,209 @@ <h2>Audit stats</h2>
...
 
-                    </tr>
-                </thead>
-                <tbody>
-                    {% for (audit, content, client_info) in audits_of_recent_content %}
-                    <tr>
-                        <td>{% if audit.trace != "" %}<a href="/audit/id/{{ audit.id }}">{{ audit.id }}</a>{% else %}
-                            {{ audit.id }}{% endif %}</td>
-                        <td><span class="badge text-bg-{% if audit.is_success() %}success{% else %}danger{% endif %}">{%
+                        </tr>
```
mrferris, line 71: Indentation/whitespace seems to have gotten a bit off in a few spots, there are plugins/tools that do the formatting for you so you don't have to worry about it, I'd search for one in your IDE. In VSCode I have a plugin that properly formats/indents any time I save.

glados-web/assets/css/content_dashboard.css
```diff
@@ -0,0 +1,11 @@
+.bootstrap-css {
```
mrferris, line 1: Let's call this something more descriptive like `content-card`

glados-web/assets/css/content_dashboard.css
```diff
@@ -0,0 +1,11 @@
...
+    border-radius: 10px;
+    width: 98% !important;
```
mrferris, line 3: Not having `!important` doesn't seem to make a difference on my browser, does it on yours? I'd think it would be only be necessary if our element was being given a conflicting width

mrferris (changes requested):
Looks great! Just a few changes/questions and this is good to go 👍

glados-web/assets/css/content_dashboard.css
mrferris: Oh, on the main PR page it doesn't show the marking but it does on the "Files Changed" page

glados-web/assets/css/content_dashboard.css
KaydenML: Ok thank you for giving this a review I'll work on fixing these changes, but for me I needed the important for sure for the width I will have to double check for the length though

mrferris (approved):

Issue #135, opened 2023-06-09
Content Dashboard loads slowly


mrferris:
http://glados.ethportal.net/content/ loads slowly.

[The endpoint](https://github.com/ethereum/glados/blob/d15e97a716deac4b301ac01a5b2e0cc733e31c47/glados-web/src/routes.rs#L205) for this page has 8 queries. The time they currently take (with 3.9 mil audit rows) are as follows:

Content ID: 120 ms
Content:      1641 ms
Audits:         451 ms
Successes: 303 ms
Failures:      377 ms
Hour stats: 941 ms
Day stats:   938 ms
Week stats 958 ms
                   = 5729 ms

The two priorities should be improving 
- [ ]  "Content" which is the most recent content that has been audited
- [ ] Stats, note that this is with #134 so despite being sped up it's still a second per stat period.

mrferris:
#119 will improve the speed of some of these via indexes, but will require wiping of DB so wanted to note load times now to be compared with once we reach this DB scale again with proper indexes.

PR #146, opened 2023-08-05
Run content dashboard queries concurrently rather than sequentially

mrferris:
Database queries for populating the content dashboard were running sequentially. 

This PR runs them concurrently.

This improves issue #135 drastically.


glados-web/src/routes.rs
```diff
@@ -214,22 +214,28 @@ pub async fn content_dashboard(
...
+    );
+
+    // Get results from queries
+    let audits_of_recent_content: Vec<AuditTuple> = results.0?;
+    let recent_audits: Vec<AuditTuple> = results.1?;
+    let recent_audit_successes: Vec<AuditTuple> = results.2?;
+    let recent_audit_failures: Vec<AuditTuple> = results.3?;
+    let hour_stats = results.4?;
+    let day_stats = results.5?;
+    let week_stats = results.6?;
```
carver, line 235: *nit* these high index numbers feel a little awkward, and an opportunity for bugs as the code evolves. I think it is a silght improvement to destructure up at the join and then self-reference to unwrap the errors, like:
```suggestion
    let audits_of_recent_content: Vec<AuditTuple> = audits_of_recent_content?;
    let recent_audits: Vec<AuditTuple> = recent_audits?;
    let recent_audit_successes: Vec<AuditTuple> = recent_audit_successes?;
    let recent_audit_failures: Vec<AuditTuple> = recent_audit_failures?;
    let hour_stats = hour_stats?;
    let day_stats = day_stats?;
    let week_stats = week_stats?;
```

That saves us from some slightly annoying bulk line changes in the future, if we need to remove, for example, the 2nd element from the joined tasks.

carver (approved):
Makes sense! 🚢

PR #151, opened 2023-08-18
Styled Content ID Page

KaydenML:
Before Image
image
After Image
image


KaydenML:
After Image, after styling using bootstrap

glados-web/templates/contentid_detail.html
```diff
@@ -2,27 +2,36 @@
 
 {% block title %}Content ID: {{ content_id.id_as_hex() }}{% endblock %}
 
+{% block head %}
+<link href="/static/css/glados_pages.css" rel="stylesheet">
```
pipermerriam, line 6: Seeing this show up in multiple pull requests suggests that *maybe* it would be cleaner to have a common base-template that each of these templates uses so that this line doesn't have to be repeated multiple times.

pipermerriam (approved):

PR #159, opened 2023-09-15
Adds embedded grafana metrics for trin node

mrferris:
Adds a `/metrics/` page that shows the message & storage metrics for glados's associated trin node, if available.

Implements #143


mrferris:
After further discussion, it would likely be better to have this viewable through a separate domain than glados.

PR #162, opened 2023-09-25
Adds monitoring and auditing of beacon data

mrferris:
This PR adds glados support for beacon data.

Changes:

- Adds `glados-monitor follow-beacon-head` command
- Modifies `glados-audit` to audit content keys from any portal overlay (previously any content key would parsed as a `HistoryContentKey` and be passed to a `portal_history...` endpoint), including state and beacon.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2429,6 +2429,7 @@ dependencies = [
  "migration",
  "reqwest",
  "sea-orm",
+ "serde_json",
  "tokio",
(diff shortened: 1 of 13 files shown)


KolbyML (approved):
:shipit:

KolbyML:
The contentkey_details.html page will crash if a beacon content_key is loaded, but I fixed the issue in my PR, so we can just merge this

PR #165, opened 2023-09-28
Adding Audit Stats To Front Page

KaydenML:
image


glados-web/src/templates.rs
```diff
@@ -12,6 +12,12 @@ use crate::routes::{PieChartResult, Stats};
 #[template(path = "index.html")]
 pub struct IndexTemplate {
     pub pie_chart_client_count: Vec<PieChartResult>,
+    pub stats: [Stats; 3],
```
mrferris, line 15: We only need these stats in the `IndexTemplate`, we don't want to pass in anything we're not going to use.

glados-web/src/routes.rs
```diff
@@ -117,8 +117,54 @@ pub async fn root(Extension(state): Extension<Arc<State>>) -> impl IntoResponse
...
+        recent_audit_successes,
+        recent_audit_failures,
+        hour_stats,
+        day_stats,
+        week_stats,
+    ) = tokio::join!(
+        get_audits_for_recent_content(KEY_COUNT, &state.database_connection),
+        get_recent_audits(KEY_COUNT, &state.database_connection),
+        get_recent_audit_successes(KEY_COUNT, &state.database_connection),
+        get_recent_audit_failures(KEY_COUNT, &state.database_connection),
```
mrferris, line 145:
```suggestion

```
Each of these calls are unnecessary because we aren't displaying them on this page.

mrferris (changes requested):

mrferris (approved):
Looks good, just missing a newline at the end of `index.html` and then i'll merge

PR #168, opened 2023-09-30
Expects utpTransfer key in content response

### LINKED ISSUE #160: Glados audit incompatibility with Fluffy - fix out of date portal JSON-RPC API?
Latest version of glados is no longer compatible with Fluffy.

Each audit returns a failure even though there is clearly data arriving:

```
[2023-09-20T12:37:11Z DEBUG hyper::proto::h1::conn] incoming body is content-length (73007 bytes)
[2023-09-20T12:37:11Z DEBUG hyper::proto::h1::conn] incoming body completed
[2023-09-20T12:37:11Z ERROR glados_audit] Problem requesting content from Portal node. content.key="0x01e57dc6f3241f3a4f8c55293a3ec3afe21563f3614c0cfa049facc7314ee2460b" err=ContainsNone
```

From some further debug it looks like the json response parsing (`as_str`) is failing: https://github.com/ethereum/glados/blob/master/glados-core/src/jsonrpc.rs#L200

There has been a change related to this to the JSON RPCall: https://github.com/ethereum/portal-network-specs/commit/92b79b8ebb67c7d29383fde489b6cfc0de5ab28d

Retesting this with a modified fluffy build that returns just the content(like before the spec change) , not the object works.

Can it be that the json-rpc portal api here needs an update?

In fluffy we implemented this change as else portal-hive was failing, so not sure how this works for Trin, perhaps due to the usage of `portal_historyTraceRecursiveFindContent`  in glados?

mrferris:
Fixes #160

Tested successfully with fluffy node.

I was tempted to finish & merge #138 in order to fix the root issue here in a more long-term way. Decided to patch this and will thaw & merge that PR subsequently.

### DIFF
--- a/glados-core/src/jsonrpc.rs
+++ b/glados-core/src/jsonrpc.rs
@@ -167,6 +167,13 @@ pub struct TracedQueryResult {
     pub trace: Value,
 }
 
+#[allow(non_snake_case)]
+#[derive(Deserialize)]
+pub struct QueryResult {
+    pub content: String,
+    pub utpTransfer: bool,
+}
+
 pub struct Content {
     pub raw: Vec<u8>,
 }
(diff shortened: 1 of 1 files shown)


mrferris:
@kdeme This should fix #160, can't formally request review from you

kdeme (approved):
Looks good!

Tested also myself locally.

PR #170, opened 2023-10-01
Client Pie Chart turns into a black circle when only one type of clie…

KaydenML:
image

### DIFF
--- a/glados-web/assets/js/piechart.js
+++ b/glados-web/assets/js/piechart.js
@@ -111,12 +111,16 @@ function pie_chart_count(pie_chart_client_count) {
         return hasData ? `${d.name}\n${d3.format(",")(d.value)}` : "";
     };
 
+    const singleClient = new Set(char_array.map(d => d.name));
+    const colors = singleClient.size > 1 ? undefined : ["red"];
+
     const chart = PieChart(char_array, {
         name: d => d.name,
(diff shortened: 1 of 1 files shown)


glados-web/assets/js/piechart.js
```diff
@@ -111,12 +111,16 @@ function pie_chart_count(pie_chart_client_count) {
         return hasData ? `${d.name}\n${d3.format(",")(d.value)}` : "";
     };
 
+    const singleClient = new Set(char_array.map(d => d.name));
```
mrferris, line 114: This doesn't seem like it should be called `singleClient`. Maybe just `clients`?

glados-web/assets/js/piechart.js
```diff
@@ -111,12 +111,16 @@ function pie_chart_count(pie_chart_client_count) {
...
+    const colors = singleClient.size > 1 ? undefined : ["red"];
```
mrferris, line 115: I'm bikeshedding, but I think that white might be a better color. That way it won't even look like a pie chart, since a pie chart with only one thing in it generally shouldn't be a pie chart anyway.

And a big red circle isn't great because the color red on the page is already associated with failure.

mrferris (approved):
Great 🚀

Just fix the two comments and this is good to go

PR #173, opened 2023-10-05
docs: update trin CLI launch docs

carver:
- One argument was counter-productive (don't turn stun off without a good reason)
- A few argements were just noisily resetting the default (see: ipc, bootnodes)
- A few arguments are no longer available (provider-related)
- Not sure why the discovery port was set to non-standard 9009, but it seems unnecessary.

The docs link was outdated and broken.

### DIFF
--- a/SETUP_GUIDE.md
+++ b/SETUP_GUIDE.md
@@ -24,15 +24,9 @@ Make an empty database file for glados.
 ```command
 $ touch /path/to/database.sqlite
 ```
-Start trin (see [docs](https://github.com/ethereum/trin/blob/master/docs/ubuntu_guide.md)
+Start trin (see [docs](https://ethereum.github.io/trin/developers/quick_setup.html))
 ```command
-~/trin$ RUST_LOG=debug cargo run -p trin -- \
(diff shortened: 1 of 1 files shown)


SETUP_GUIDE.md
```diff
@@ -24,15 +24,9 @@ Make an empty database file for glados.
 ```command
 $ touch /path/to/database.sqlite
 ```
-Start trin (see [docs](https://github.com/ethereum/trin/blob/master/docs/ubuntu_guide.md)
+Start trin (see [docs](https://ethereum.github.io/trin/developers/quick_setup.html))
 ```command
-~/trin$ RUST_LOG=debug cargo run -p trin -- \
-    --no-stun \
-    --web3-transport ipc \
```
KolbyML, line 31: most glados utilities use http instead of ipc which is the default. It might be good to leave ``--web-transport ipc`` here to make it easier to switch between them if people are copy pasting. Or just use ``--web-transport http`` since most glados utilities only support http

mrferris (approved):

PR #175, opened 2023-10-07
Adds hourly success rate API route for glados bot integration

mrferris:
Adds glados API route `/api/hourly_success_rate` for the purpose of integrating a discord bot to alert us when success rate drops.

Returns just a single JSON float value representing the percentage of success, or an error message on error if one occurs.

### DIFF
--- a/glados-web/src/lib.rs
+++ b/glados-web/src/lib.rs
@@ -70,6 +70,10 @@ pub async fn run_glados_web(config: Arc<State>) -> Result<()> {
             get(routes::contentkey_detail),
         )
         .route("/audit/id/:audit_id", get(routes::contentaudit_detail))
+        .route(
+            "/api/hourly-success-rate/",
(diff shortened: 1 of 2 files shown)


KolbyML (approved):

glados-web/src/lib.rs
```diff
@@ -70,6 +70,10 @@ pub async fn run_glados_web(config: Arc<State>) -> Result<()> {
             get(routes::contentkey_detail),
         )
         .route("/audit/id/:audit_id", get(routes::contentaudit_detail))
+        .route(
+            "/api/hourly_success_rate/",
```
pipermerriam, line 74: -0 on the simple aesthetics of underscores in a URL.  Personal preference but I much prefer dashes:

- `/api/hourly-success-rate/`

mrferris: true, URLs get underlined often and that looks weird with underscores

PR #176, opened 2023-10-07
Radius Node Id scatter chart

KaydenML:


KaydenML:
image
Picture Of The Graph Currently, Trying To Get The Labels To Look Nice And Not Overlap.

KaydenML:
image
When using a bar chart like this and displaying the data the bars appear so small that you can't even see them

image
But If i increase the width of the bars they overlap over each other, I could round the data up but it would give a less accurate representation of the data.

While using a density chart I was having problems displaying the data in a way that looks nice

pipermerriam:
What about:

- horizontal axis is node-id
- vertial axis is representation of radius (percentage?)

All nodes are represented on the graph as a dot?

In theory, you can then scan to see if the node-id keyspace has any weird hotspots with weird clustering of radius values and scanning the vertical axis shows normative radius values....  

I am betting that there's some manual scaling of the graph that would be needed since most of the values will be very small, and then there will be a few 100% ones that throw the scale way off.

KaydenML:
Thank you piper

KaydenML:
I'll begin looking into using this type of graph I really appreciate your time!

pipermerriam:
Feel free to move forward with this however you find to best represent it.  I would encourage you to use this "litmus test" on whether the graph is useful.

- Can we clearly see the clustering and overall distribution of node radius values of nodes in the network.

KaydenML:
image

KaydenML:
@pipermerriam Hey Piper it's ready to review now

glados-web/assets/css/homepage.css
```diff
@@ -48,3 +48,9 @@ h1 {
     box-shadow: 0 7px 25px rgba(0, 0, 0, 0.08);
     height: 325px;
 }
+
```
pipermerriam, line 51: Might be nice to "future" selves to add a comment to this that indicates why it is here and what it is for.  Over time, the incremental accumulation of CSS styles will grow...

glados-web/src/routes.rs
```diff
@@ -118,6 +131,77 @@ pub async fn root(Extension(state): Extension<Arc<State>>) -> impl IntoResponse
...
+            Alias::new("data_radius"),
+        )
+        .expr(Expr::col((node::Entity, node::Column::NodeId)))
+        .from(census_node::Entity)
+        .from(node::Entity)
+        .from(record::Entity)
+        .from_subquery(
+            Query::select()
+                .from(census::Entity)
+                .expr_as(Expr::max(Expr::col(census::Column::Id)), Alias::new("id"))
```
pipermerriam, line 147: Using the auto-incrementing property of the `id` column when we have a `started_at` column that actually provides chronological ordering seems like something that might be worth the effort to do *correctly*.

I will acknowledge that the two approaches are functionally equivalent at this stage of the property, but if we were to move towards UUIDs or other identifiers in the future (which we're unlikely to do) this would break...  

So at a minimum, this might justify a comment acknowledging this situation.

glados-web/templates/index.html
```diff
@@ -59,11 +61,19 @@ <h3 style="text-align: center">Client Count</h3>
             </div>
         </div>
     </div>
+    <br>
+    <div class="card shadow-sm card-box">
+        <div class="card-body">
+            <div id="my_dataviz"></div>
+        </div>
+    </div>
+    <div id="hover" style="position: absolute; opacity: 0;" ></div>
```
mrferris, line 70:
```suggestion
    <div id="hover" style="pointer-events: none; position: absolute; opacity: 0;" ></div>
```

This will fix the bug I'm seeing where a node's hover-over doesn't work if it was underneath the previous hover-over

mrferris: I would also put a bit of a background on this div, maybe a bit see-through, so that the details are easier to read

glados-web/assets/js/radiusdensity.js
```diff
@@ -0,0 +1,72 @@
+function radius_node_id_scatter_chart(data) {
+    const margin = {top: 20, right: 25, bottom: 50, left: 25},
+        width = 450 - margin.left - margin.right,
+        height = 400 - margin.top - margin.bottom;
+
+// append the svg object to the body of the page
```
mrferris, line 6: nit: indent comments in this file

glados-web/templates/index.html
```diff
@@ -59,11 +61,19 @@ <h3 style="text-align: center">Client Count</h3>
             </div>
         </div>
     </div>
+    <br>
+    <div class="card shadow-sm card-box">
+        <div class="card-body">
+            <div id="my_dataviz"></div>
```
mrferris, line 67: make sure you change placeholders

`census-scatterplot` would be more appropriate

glados-web/templates/base.html
```diff
@@ -9,6 +9,7 @@
     <script src="https://code.jquery.com/jquery-3.6.1.min.js"></script>
     <link rel="stylesheet" href="/static/css/bootstrap.min.css">
     <link rel="stylesheet" href="/static/css/trace.css">
+    <link href="/static/css/glados_pages.css" rel="stylesheet">
```
mrferris, line 12:
```suggestion
    <link href="/static/css/glados_styles.css" rel="stylesheet">
```

glados-web/assets/css/homepage.css
```diff
@@ -48,3 +48,9 @@ h1 {
...
+.card-box {
```
mrferris, line 52: Why is `.card-box` being used by the scatterplot but not by the stats and pie chart on the same page? The card borders and everything look the same to me

glados-web/src/routes.rs
```diff
@@ -40,6 +40,19 @@ pub async fn handle_error(_err: io::Error) -> impl IntoResponse {
     (StatusCode::INTERNAL_SERVER_ERROR, "Something went wrong...")
 }
 
+#[derive(FromQueryResult, Serialize, Debug)]
+pub struct RadiusChartData {
+    pub data_radius: Vec<u8>,
+    pub node_id: Vec<u8>,
+}
+
+#[derive(FromQueryResult, Serialize, Debug)]
```
mrferris, line 49: It doesn't seem like this needs the `FromQueryResult` trait

glados-web/src/routes.rs
```diff
@@ -40,6 +40,19 @@ pub async fn handle_error(_err: io::Error) -> impl IntoResponse {
     (StatusCode::INTERNAL_SERVER_ERROR, "Something went wrong...")
 }
 
+#[derive(FromQueryResult, Serialize, Debug)]
```
mrferris, line 43: It doesn't seem like this needs the `FromQueryResult` trait

glados-web/src/routes.rs
```diff
@@ -118,6 +131,77 @@ pub async fn root(Extension(state): Extension<Arc<State>>) -> impl IntoResponse
         .await
         .unwrap();
 
+    let mut average_radius = Query::select();
```
mrferris, line 134: this entire thing should also be wrapped in a function that returns a `Vec<CalculatedRadiusChartData>` that is called inside of the main `root` function instead of being inline

glados-web/src/routes.rs
mrferris: Yea I think doing it based off `started_at` is better in principle

mrferris (changes requested):

KaydenML:
https://github.com/ethereum/glados/pull/180 
kolby's code in this PR and his usage of started_at was a really big help in fixing this change.

glados-web/src/routes.rs
```diff
@@ -118,6 +131,77 @@ pub async fn root(Extension(state): Extension<Arc<State>>) -> impl IntoResponse
...
+    average_radius
+        .expr_as(
+            Expr::col(census_node::Column::DataRadius),
```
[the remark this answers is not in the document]
mrferris: for posterity: kayden pointed out that this approach won't allow the entire node ID to be displayable

mrferris (approved):

PR #177, opened 2023-10-09
allow trace requests for fluffy

mynameisdaniil:
Allow trace feature for Fluffy and (potentially) other clients.
+ removed unused fields
+ all `snake_case` fields renamed to `camelCase`
+ ENR decoder for port and ip
+ support for displaying cancelled requests

Here's spec update: https://github.com/ethereum/portal-network-specs/pull/236


mrferris (approved):
🫡

PR #179, opened 2023-10-16
Adds humanized timestamps to dashboard with full timestamps on hoverover

mrferris:


entity/src/utils.rs
```diff
@@ -0,0 +1,39 @@
+use chrono::{DateTime, FixedOffset, Utc};
+use std::fmt::Write;
+
+/// Returns a string representing the time elapsed since the given time.
+/// eg "2m1w2d3h4m5s ago"
+pub fn time_ago(past: DateTime<FixedOffset>) -> String {
```
pipermerriam, line 6: This kind of method is easier to test if it allows `now` to be computed externally and passed into it....

pipermerriam: Alternatively, two-layer function, one that allows `now` to be passed in externally, and then a simple wrapper that computes `now` automatically.

PR #180, opened 2023-10-19
feat: api to check if content is in deadzone or will be accepted by network

KolbyML:
The SQL which is most of this PR is just copy pasted from @KaydenML draft-PR so I am co-authoring the commit https://github.com/ethereum/glados/pull/176

There has been talks about possible deadzones on our network. But I didn't really think of it much till I ran the bridge in node count 16 mode and tried to gossip a block body 20 times or something. This is when I came to the realization what Mike and I believe Nick mentioned to check... a deadzone. I had an assumption that with the radius's on the network that couldn't be an issue, but after all maybe it could be a deadzone?

So this api checks if any nodes found in the latest census would accept the content_key if they were offered.

### DIFF
--- a/glados-web/src/routes.rs
+++ b/glados-web/src/routes.rs
@@ -11,6 +11,8 @@ use entity::{
     content_audit::{self, AuditResult},
     execution_metadata, key_value, node, record,
 };
+use ethportal_api::jsonrpsee::core::__reexports::serde_json;
+use ethportal_api::types::distance::{Distance, Metric, XorMetric};
 use ethportal_api::utils::bytes::{hex_decode, hex_encode};
 use ethportal_api::{HistoryContentKey, OverlayContentKey};
 use migration::{Alias, JoinType};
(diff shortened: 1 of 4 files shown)


KolbyML:
That is very true. Currently the code only supports historical census's it looks like so I can't specify that till multi-network census support is implemented.

> It seems that we won't always want to see this against the latest census, but rather see it against a specific census at a specific point in time...

That will probably be a more general approach futher down the line. I am looking to add this kind of as a debugging tool for the live network. Then an historical census explorer, but that functionality can be quicky added when/if a historical census explorer is implemented.

pipermerriam:
I'm :+1: on you moving forward with this as-is, I did make an issue for the broader functionality to be added.

glados-web/src/routes.rs
```diff
@@ -689,6 +691,69 @@ impl Period {
...
+        )
+        .expr(Expr::col((node::Entity, node::Column::NodeId)))
+        .expr(Expr::col((record::Entity, record::Column::Raw)))
+        .from(census_node::Entity)
+        .from(node::Entity)
+        .from(record::Entity)
+        .from_subquery(
+            Query::select()
+                .from(census::Entity)
+                .expr_as(Expr::max(Expr::col(census::Column::Id)), Alias::new("id"))
```
pipermerriam, line 720: I believe that this would be *more correct* if it queried against the `started_at` timestamp.  Likely to remain functionally equivalent for the indefinite future, but I tend to find this as a sort of "code smell" since there's no actual guarantee that the highest ID in the database will correspond to the latest census...  where-as the timestamp is there for exactly that purpose.

If you want bonus points, you can also take into account the `duration` field...

glados-web/src/routes.rs
```diff
@@ -689,6 +691,69 @@ impl Period {
...
+    let dead_zone_data = DeadZoneData::find_by_statement(builder.build(&average_radius))
+        .all(&state.database_connection)
+        .await
+        .unwrap();
+
+    let content_key: ethportal_api::HistoryContentKey =
+        serde_json::from_value(serde_json::json!(content_key)).unwrap();
+
+    let mut enrs: Vec<String> = vec![];
+    for i in dead_zone_data {
```
pipermerriam, line 746: what is `i`?  Maybe use a more descriptive variable name?

glados-web/src/routes.rs
```diff
@@ -689,6 +691,69 @@ impl Period {
...
+        let radius = Distance::from(crate::U256::from_big_endian(&i.data_radius));
+        let node_id = Distance::from(crate::U256::from_big_endian(&i.node_id));
+        if XorMetric::distance(&content_key.content_id(), &node_id.big_endian()) <= radius {
```
pipermerriam, line 749: seems slightly wasteful to recompute the `content_id` at every iteration of this loop.

glados-web/src/routes.rs
KolbyML: I updated this let me know if  you like it

glados-web/src/routes.rs
KolbyML: resolved

glados-web/src/routes.rs
KolbyML: Sounds good I also took duration into account so that is cool

KolbyML:
@pipermerriam I believe I resolved your requests so you can take another look at it now

mrferris (approved):

PR #156, opened 2023-09-08
Implemented A Segmental Control Feature That Only Displays The Table You Toggle To

KaydenML:
After Images
image
image
image

PR #183, opened 2023-10-21
feat: Adds audit dashboard with filters

### LINKED ISSUE #158: Filter audit data & stats by portal data type

mrferris:
This pr adds audit dashboard, which is similar to the current content dashboard except that it shows one pane of audits and allows them and their hour/day/week stats to be filtered by any combination of strategy, content type, and success. 

Similar idea as #156, except that only the data required is fetched from the server via an `/audits/filter/` endpoint, rather than all of the data being loaded and most of it being hidden.

closes #158 and #182

### DIFF
--- a/glados-web/src/routes.rs
+++ b/glados-web/src/routes.rs
@@ -1,11 +1,11 @@
 use axum::{
-    extract::{Extension, Path},
+    extract::{Extension, Path, Query as HttpQuery},
     http::StatusCode,
     response::IntoResponse,
     Json,
 };
 use chrono::{DateTime, Duration, Utc};
-use entity::{census, census_node, client_info};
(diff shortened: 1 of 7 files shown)


glados-web/src/routes.rs
```diff
@@ -700,6 +826,59 @@ pub struct Stats {
     pub audits_per_minute: u32,
 }
 
+async fn get_filtered_audit_stats(
```
pipermerriam, line 829: This method looks "testable", and while a little tedious to write the tests, it seems like the kind of thing that should/could be tested to increase our confidence that these filters are indeed implemented correctly both now, and across future refactoring.

mrferris: Agreed, will add in a separate PR rather after I remove sqlite from CI, because this uses postgres primitives

PR #186, opened 2023-10-28
Fixes `glados-cartographer` showing few trin nodes

### LINKED ISSUE #141: Cannot create Non-Unique indices using SeaORM
There are a few SQL indices that we want for performance purposes that we do not want to have `UNIQUE` constraints:

(`content_audit::CreatedAt`, `content_audit::Result`)
(`content::FirstAvailableAt`, `content::ProtocolId`)
key_value::Key
execution_metadata::BlockNumber

Creating a non-unique index in `seaORM` via eg: 

```
 .index(
         Index::create()
        .name("idx_execution-block_number")
       .col(ExecutionMetadata::BlockNumber),
)
```

results in the error:

```
thread 'main' panicked at 'Database migration failed: Exec(SqlxError(Database(PgDatabaseError { severity: Error, code: "42601", message: "syntax error at or near \"(\"", detail: None, hint: None, position: Some(Original(172)), where: None, schema: None, table: None, column: None, data_type: None, constraint: None, file: Some("scan.l"), line: Some(1188), routine: Some("scanner_yyerror") })))', glados-monitor/src/main.rs:47:14
```

The offending SQL appears to be:

```
CREATE TABLE "execution_metadata" (
      "id" serial NOT NULL PRIMARY KEY,
      "content" integer NOT NULL,
      "block_number" integer NOT NULL,
      CONSTRAINT "idx_execution-block_number" ("block_number"),
      CONSTRAINT "idx-unique-metadata" UNIQUE ("content"),
      CONSTRAINT "FK_executionmetadata_content" FOREIGN KEY ("content") REFERENCES "content" ("id") ON DELETE
      SET
        NULL ON
      UPDATE
        CASCADE
    )
```

while the working SQL generated with a unique constraint looks like:

```
CREATE TABLE "execution_metadata" (
      "id" serial NOT NULL PRIMARY KEY,
      "content" integer NOT NULL,
      "block_number" integer NOT NULL,
      CONSTRAINT "idx_execution-block_number" UNIQUE ("block_number"),
      CONSTRAINT "idx-unique-metadata" UNIQUE ("content"),
      CONSTRAINT "FK_executionmetadata_content" FOREIGN KEY ("content") REFERENCES "content" ("id") ON DELETE
      SET
        NULL ON
      UPDATE
        CASCADE
    )
```

The only difference is the `UNIQUE` on the fifth line, so how that's resulting in a `"syntax error at or near \"(\""` is unclear.

mrferris:
The `record` entity's `(nodeId, seqNo)` index was defined as unique, so new records for the same node weren't getting added during census.

What changed:

- Fixed #141 (by defining the index separately from the table) and made the index non-unique. 
- All missing unique & non-unique indices discussed in [this PR](https://github.com/ethereum/glados/pull/119) were added.
- Timestamps were switched to `DateTime<Utc>` instead of `DateTime<FixedOffset>` for best handling of daylight savings time etc.


pipermerriam (commented):
Some confusion around this being implemented by modifying migrations that have already been run.  What is the deployment plan for this?

pipermerriam (approved):
Code all looks fine, and as long as you have a deployment plan for how to get these migrations run such that they add the indexes, I'm :+1: on moving this forward.

mrferris:
for posterity, from discord: 
> if it were just indices I'd have done a new migration file but with the timestamp changes (not a design change, just fixing a pretty objective mistake that I'd made prior) I was planning on just resetting the DB and re-applying migrations, and then aim to keep things persistent from here on out

PR #189, opened 2023-11-03
Makes metadata block number index non-unique

mrferris:
The index for execution metadata's block number column was made unique, which causes some problems because currently `Content` has a one-to-one relationship with `ExecutionMetadata`, where the index was assuming many-to-one. This fixes that.

### DIFF
--- a/migration/src/m20230511_104838_create_execution_metadata.rs
+++ b/migration/src/m20230511_104838_create_execution_metadata.rs
@@ -6,7 +6,7 @@ pub struct Migration;
 #[async_trait::async_trait]
 impl MigrationTrait for Migration {
     async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
-        manager
+        let _ = manager
             .create_table(
                 Table::create()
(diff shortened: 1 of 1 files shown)


migration/src/m20230511_104838_create_execution_metadata.rs
```diff
@@ -49,6 +43,15 @@ impl MigrationTrait for Migration {
                     )
                     .to_owned(),
             )
+            .await;
+        manager
+            .create_index(
+                Index::create()
+                    .name("idx_metadata-blocknumber")
```
njgheorghita, line 50: noticed that this name (`executionmetdata` -> `metadata` & `block_number` -> `blocknumber`) is different from before, does that matter?

njgheorghita (approved):
🚢

migration/src/m20230511_104838_create_execution_metadata.rs
mrferris: nope, but i changed it back anyway

PR #191, opened 2023-11-07
feat: adds audit stats recording and historical display

### LINKED ISSUE #164: A way to view audit metrics over time
Glados is good at showing us how well the network is doing right now.

I think the main index/dashboard should be updated so that it provides context for how well we are doing with respect to how well we previously were doing.  A simple view of this might be showing a digest for the last month, with each day segmented off and showing aggregate success fail metrics.

- This week 70% success
- Last week 60% success
- The week before that 62% success

The current "last hour", "last day" and "last week" metrics are still good, but I think they become less interesting/important than looking at how well we've been doing over longer time arcs.

mrferris:
This PR resolves #164, with a chart instead of the day segments. This aggregate data can be added in a subsequent PR.

The UI is a little rough around the edges, but good enough to be usable. 

The only question on the data gathered is whether I should continue calculating it as the average over the past hour. Maybe doing the average over a smaller period would be better. We've been using the "average over the past hour" metric when discussing glados performance, so that's what I went with.

Changes:

- Adds the `audit_stats` table, where stats for a bunch of different combinations of filters are recorded
- Moves stat calculation logic from `glados-web` to `glados-core` so that both `glados-web` and `glados-audit` can use 
- Adds the `periodically_record_stats` task to `glados-audit`
- Adds the `/api/stat-history/` endpoint
- Adds `stats_history.js` widget, and instantiates it on the index page after requesting the data from the endpoint
- Adds the `--stats-recording-period` flag to `glados-audit`, which defaults to 300 seconds

For local testing purposes I would set the recording period to like 5 seconds so that you get test data reasonably quickly.

Screen Shot 2023-11-07 at 2 56 46 PM

### DIFF
--- a/entity/src/audit_stats.rs
+++ b/entity/src/audit_stats.rs
@@ -1,0 +1,84 @@
+use chrono::Duration;
+
+use anyhow::Result;
+use chrono::{DateTime, Utc};
+use sea_orm::{entity::prelude::*, ActiveValue::NotSet, QueryOrder, Set};
+use serde::Serialize;
+
+#[derive(Clone, Debug, DeriveEntityModel, Serialize)]
+#[sea_orm(table_name = "audit_stats")]
+pub struct Model {
+    #[sea_orm(primary_key)]
+    pub id: i32,
+    pub timestamp: DateTime<Utc>,
+    pub num_audits: i32,
+    pub success_rate_all: f32,
+    pub success_rate_latest: f32,
+    pub success_rate_random: f32,
+    pub success_rate_oldest: f32,
+    pub success_rate_all_headers: f32,
+    pub success_rate_all_bodies: f32,
+    pub success_rate_all_receipts: f32,
+    pub success_rate_latest_headers: f32,
+    pub success_rate_latest_bodies: f32,
+    pub success_rate_latest_receipts: f32,
+    pub success_rate_random_headers: f32,
+    pub success_rate_random_bodies: f32,
+    pub success_rate_random_receipts: f32,
+}
+
+#[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
+pub enum Relation {}
+
+impl ActiveModelBehavior for ActiveModel {}
+
+#[allow(clippy::too_many_arguments)]
+pub async fn create(
+    timestamp: DateTime<Utc>,
+    num_audits: i32,
+    success_rate_all: f32,
+    success_rate_latest: f32,
+    success_rate_random: f32,
+    success_rate_oldest: f32,
+    success_rate_all_headers: f32,
+    success_rate_all_bodies: f32,
+    success_rate_all_receipts: f32,
+    success_rate_latest_headers: f32,
+    success_rate_latest_bodies: f32,
+    success_rate_latest_receipts: f32,
+    success_rate_random_headers: f32,
+    success_rate_random_bodies: f32,
+    success_rate_random_receipts: f32,
+    conn: &DatabaseConnection,
+) -> Result<Model> {
+    let audit_stats = ActiveModel {
+        id: NotSet,
+        timestamp: Set(timestamp),
+        num_audits: Set(num_audits),
+        success_rate_all: Set(success_rate_all),
+        success_rate_latest: Set(success_rate_latest),
+        success_rate_random: Set(success_rate_random),
+        success_rate_oldest: Set(success_rate_oldest),
+        success_rate_all_headers: Set(success_rate_all_headers),
+        success_rate_all_bodies: Set(success_rate_all_bodies),
+        success_rate_all_receipts: Set(success_rate_all_receipts),
+        success_rate_latest_headers: Set(success_rate_latest_headers),
+        success_rate_latest_bodies: Set(success_rate_latest_bodies),
+        success_rate_latest_receipts: Set(success_rate_latest_receipts),
+        success_rate_random_headers: Set(success_rate_random_headers),
+        success_rate_random_bodies: Set(success_rate_random_bodies),
+        success_rate_random_receipts: Set(success_rate_random_receipts),
+    };
+    Ok(audit_stats.insert(conn).await?)
+}
+
+/// Get the most recent audit stat series of the last 7 days.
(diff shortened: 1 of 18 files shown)


pipermerriam (commented):
I'm -0 on the table based approach for this... is there a reason we can't compute these on-the-fly?  are the queries too expensive?  if the queries end up being too expensive can we cache the results of the queries to amortize this cost....

The table feels like a "poor-man's cache" and has a small code-smell to me.  thoughts?

mrferris:
Yea my thinking was that we want the data to remain "explorable". Caching an expensive calculation works until you change the time frame, at which point the cost needs to be incurred again. For example, zooming in to get better resolution on a specific time frame, zooming out to get a bigger picture view.

I assumed that doing the calculations for one metric on-the-fly would be doable, but that doing it for the 13 metrics displayed above would be too much. That said, I will get some more concrete numbers on the on-the-fly approach.

glados-web/templates/index.html
```diff
@@ -62,18 +63,28 @@ <h3 style="text-align: center">Client Count</h3>
...
+                <div class="card-body">
+                    <div id="census-scatterplot"></div>
+                </div>
+            </div>
+            <div id="hover" style="pointer-events: none; position: absolute; opacity: 0; z-index: 100;"></div>
+            <div class="card shadow-sm card-box col" id="stats-graph-container">
+                <div class="card-body">
+                    <div id="stats-history-graph"> </div>
+                </div>
+            </div>
```
KolbyML, line 78:
```suggestion
        <div class="col">
            <div class="card shadow-sm card-box col">
                <div class="card-body">
                    <div id="census-scatterplot"></div>
                </div>
            </div>
            <div id="hover" style="pointer-events: none; position: absolute; opacity: 0; z-index: 100;"></div>
            </div>
            
            <div class="col">
            <div class="card shadow-sm card-box col" id="stats-graph-container">
                <div class="card-body">
                    <div id="stats-history-graph"> </div>
                </div>
            </div>
```

entity/src/audit_stats.rs
```diff
@@ -0,0 +1,84 @@
+use chrono::Duration;
+
+use anyhow::Result;
+use chrono::{DateTime, Utc};
+use sea_orm::{entity::prelude::*, ActiveValue::NotSet, QueryOrder, Set};
```
KolbyML, line 5: I think I heard someone on the team say we should avoid using wild card imports, could have misremembered that though.

KolbyML (commented):
Mike asked me to take a look at it earlier today. I just found 2 issues, 1 which would obviously be handled in a future PR.

I don't fully know the vision of Glados or if a graph should be or not be included or not, so I am just commenting objectively on the change diff and testing if the PR works as expected locally. So far it looks to be working as expected, but I will update this message in 3ish hours if I see otherwise.

So I guess what I am saying is that I just quickly reviewed if what is here works and I am not commenting on what the future function of this PR should be or if something should be computed on the fly or not.

glados-audit/src/cli.rs
```diff
@@ -40,6 +40,8 @@ pub struct Args {
         help = "relative weight of the 'random' strategy"
     )]
     pub random_strategy_weight: u8,
+    #[arg(long, default_value = "300", help = "stats recording period (seconds)")]
```
njgheorghita, line 43: *nitpick* create a const to be used here and in the default impl for the 300 value

glados-audit/src/stats.rs
```diff
@@ -0,0 +1,204 @@
...
+/// Loops indefinitely, periodically recording audit stats to the database.
+pub async fn periodically_record_stats(period: Duration, conn: DatabaseConnection) -> ! {
+    debug!("initializing task for logging audit stats");
+    let mut interval = interval(period);
+
+    loop {
+        interval.tick().await;
+
+        record_current_stats(&conn).await.unwrap_or_else(|e| {
+            error!("failed to record audit stats: {}", e);
```
njgheorghita, line 20: *nitpick* `error!("failed to record audit stats: {e}")`

glados-audit/src/stats.rs
```diff
@@ -0,0 +1,204 @@
...
+use tokio::time::{interval, Duration};
+use tracing::{debug, error, info};
+
+/// Loops indefinitely, periodically recording audit stats to the database.
+pub async fn periodically_record_stats(period: Duration, conn: DatabaseConnection) -> ! {
+    debug!("initializing task for logging audit stats");
+    let mut interval = interval(period);
+
+    loop {
+        interval.tick().await;
```
njgheorghita, line 17: *nitpick* typically I'd place these at the end of the loop rather than the beginning, to avoid the delay before the initial calculation, but it doesn't seem like a big deal in this case

glados-audit/src/stats.rs
```diff
@@ -0,0 +1,204 @@
...
+        success_rate_latest_receipts,
+        success_rate_random_headers,
+        success_rate_random_bodies,
+        success_rate_random_receipts,
+        conn,
+    )
+    .await
+    {
+        Ok(_) => debug!("successfully recorded audit stats"),
+        Err(e) => error!("failed to record audit stats: {}", e),
```
njgheorghita, line 201: *nit* put the `e` inside `{}`

glados-core/src/stats.rs
```diff
@@ -0,0 +1,218 @@
...
+        total_audits,
+        total_passes,
+        pass_percent,
+        total_failures,
+        fail_percent,
+        audits_per_minute,
+    })
+}
+
+pub struct Stats {
```
njgheorghita, line 143: *nitpick* I'm not super intimate with the codebase, but the name `Stats` does seem a bit vague imo, especially in a codebase that's dealing with a lot of numbers / tables / data. At the very least, I think some docstrings on these types would be useful.

glados-web/assets/js/stats_history.js
```diff
@@ -0,0 +1,166 @@
+function createMultiLineChart(height, width, dataSets) {
```
njgheorghita, line 1: *nitpick* this doesn't look like valid rust to me? surprised the compiler didn't yell at you 😅

njgheorghita (approved):
Nice! The table looks neat, and can't wait to get it up on production. Nothing much of significance to add just a few nitpicks

Issue #193, opened 2023-11-09
Improvements to audit page


KolbyML:
![image](https://github.com/ethereum/glados/assets/31669092/2364f9e3-9d6d-4916-a715-193a992c4fc5)

It would be nice if
- it told you what kind of audit it was

Issue #192, opened 2023-11-09
Improvements to content key page


KolbyML:
![image](https://github.com/ethereum/glados/assets/31669092/2bafaa00-fc36-429f-94d6-20bec58373ed)

On this page it would be nice if you could 
- click content id to go to the page for that
- if the Content audits were clickable for their respective audit pages
- if the content audits told you what kind they were (latest, etc etc) instead of having to go to the audit page to check

PR #195, opened 2023-11-09
fix: add strategy to the audit page & content key page

mrferris:
Fixes everything mentioned in #192 and #193

### DIFF
--- a/glados-web/templates/contentaudit_detail.html
+++ b/glados-web/templates/contentaudit_detail.html
@@ -11,7 +11,6 @@
 
 <div id="top-of-audit-detail">
     <div class="row">
-        <h1>Audit #{{ audit.id }}</h1>
         <div class="col">
             <ul>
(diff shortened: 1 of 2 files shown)


glados-web/templates/contentaudit_detail.html
```diff
@@ -11,7 +11,6 @@
 
 <div id="top-of-audit-detail">
     <div class="row">
-        <h1>Audit #{{ audit.id }}</h1>
```
KolbyML, line 14: Were we supposed to remove this?

KolbyML (approved):
:shipit:

glados-web/templates/contentaudit_detail.html
mrferris: Yea, all this text was taking up too much room, and this header was taking up most of it, when the audit ID is in the url if we need it

PR #196, opened 2023-11-09
fix: adds audit rate calculation to audit dashboard

mrferris:
`get_filtered_audit_stats` was mistakenly setting the audit rate to 0


njgheorghita (approved):
🚢

PR #197, opened 2023-11-24
chore: update setup for postgres + added cartographer

KolbyML:
We are removing sqlite support so transitioning the setup docs to postgres.

I also changed everything so it uses http instead of ipc, as it is a pain to switch and being able to copy paste from here is nice. Most glados programs don't support ipc anyways.


KolbyML:
ty mike

Issue #163, opened 2023-09-26
Web view for DHT census data


pipermerriam:
We should have a web view to explore census data.

- Paginated list of all past census data showing (Id/CreatedAt/NumNodes)
- Detail page for a census that shows meta stats (CreatedAt/NumNodes), client diversity graph, and a list of all found ENR records

KolbyML:
I am going to pick this up, as Piper said it is a good thing to have before the feature I want. I believe Mike is focused on other area's right now and this issue should be very quick. I feel a little weird taking over a issue someone self assigned themselves to so hopefully it is alright, mike ping me if you are almost done or wanted to do this.

PR #198, opened 2023-11-24
feat: monitor the health of node routing tables

KolbyML:

PR #199, opened 2023-11-25
feat: implement web view for DHT census data

KolbyML:
fixes: https://github.com/ethereum/glados/issues/163

I implemented it so it looks pretty decent on mobile not 100% but nice to have.

image
image
image
image

Here are some pictures

### DIFF
--- a/glados-web/src/lib.rs
+++ b/glados-web/src/lib.rs
@@ -59,11 +59,8 @@ pub async fn run_glados_web(config: Arc<State>) -> Result<()> {
     // setup router
     let app = Router::new()
         .route("/", get(routes::root))
-        .route(
-            "/census-explorer/paginated-census-list/:list_census_page_id/",
-            get(routes::census_explorer_list),
-        )
-        .route("/census-explorer/:census_id/", get(routes::census_explorer))
+        .route("/census/census-list/", get(routes::census_explorer_list))
+        .route("/census/", get(routes::census_explorer))
         .route("/network/", get(routes::network_dashboard))
         .route("/network/node/:node_id_hex/", get(routes::node_detail))
         .route(
--- a/glados-web/src/lib.rs
+++ b/glados-web/src/lib.rs
@@ -59,6 +59,11 @@ pub async fn run_glados_web(config: Arc<State>) -> Result<()> {
     // setup router
     let app = Router::new()
         .route("/", get(routes::root))
+        .route(
+            "/census-explorer/paginated-census-list/:list_census_page_id/",
+            get(routes::census_explorer_list),
+        )
+        .route("/census-explorer/:census_id/", get(routes::census_explorer))
         .route("/network/", get(routes::network_dashboard))
         .route("/network/node/:node_id_hex/", get(routes::node_detail))
         .route(
--- a/glados-web/templates/census_explorer_page.html
+++ b/glados-web/templates/census_explorer_page.html
@@ -4,7 +4,6 @@
 
 {% block head %}
 <script src="/static/js/d3.min.js"></script>
-<script src="https://d3js.org/d3.v7.min.js"></script>
 <script src="/static/js/piechart.js"></script>
 <script src="/static/js/radiusdensity.js"></script>
 <link href="/static/css/homepage.css" rel="stylesheet">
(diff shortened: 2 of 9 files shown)


glados-web/templates/census_explorer_page.html
```diff
@@ -0,0 +1,106 @@
...
+                        <table class="table">
+                            <thead>
+                            <tr>
+                                <th scope="col">ENR</th>
+                            </tr>
+                            </thead>
+                            <tbody>
+                            {% for enr in enr_list %}
+                            <tr>
+                                <td>{{ enr.raw }}</td>
```
pipermerriam, line 90: might be nice to have these link to the actual ENR detail page for ENR records.

glados-web/src/lib.rs
```diff
@@ -59,6 +59,11 @@ pub async fn run_glados_web(config: Arc<State>) -> Result<()> {
     // setup router
     let app = Router::new()
         .route("/", get(routes::root))
+        .route(
+            "/census-explorer/paginated-census-list/:list_census_page_id/",
```
pipermerriam, line 63: Typically page numbers are done as URL query parameters rather than being part of the URL, aka:

`/census-explorer/census-list/?page=1`

glados-web/src/lib.rs
```diff
@@ -59,6 +59,11 @@ pub async fn run_glados_web(config: Arc<State>) -> Result<()> {
...
+            get(routes::census_explorer_list),
+        )
+        .route("/census-explorer/:census_id/", get(routes::census_explorer))
```
pipermerriam, line 66: I've got a mild preferences towards this URL being `/census/:census_id/`

glados-web/src/lib.rs
pipermerriam: I also have a mild preference that this URL be simply `/census/`

glados-web/templates/census_explorer_page.html
```diff
@@ -0,0 +1,106 @@
+{% extends "base.html" %}
+
+{% block title %}Glados{% endblock %}
+
+{% block head %}
+<script src="/static/js/d3.min.js"></script>
+<script src="https://d3js.org/d3.v7.min.js"></script>
```
pipermerriam, line 7: I'm generally opposed to pulling web assets this way.  This allows the `d3js.org` domain owners to inject any javascript they want into our application.  We should vendor the JS file locally and serve it ourselves.

pipermerriam (approved):
:+1:

glados-web/templates/census_explorer_page.html
KolbyML: That will be in my next PR. I want to rewrite the ENR detail's page. But I want this to link to a census snapshot of the routing table with all of the ENR's details.

I also want to list stats on what ENR's in the routing table weren't in the census.

KolbyML: If we know what nodes routing table look like we know if it is health

glados-web/templates/census_explorer_page.html
KolbyML: Oh weird I just copy pasted index.html I will remove this from that file too :+1:

glados-web/templates/census_explorer_page.html
KolbyML: I want to write this to visualize an issue mike mentioned on when cartographer was running nodes were returning month old offline nodes. So each census was checking 1800 enrs when only 160ish are on the network which means there is a bug, since clients should gossip enr's of nodes offline non-connected nodes. For context*

pipermerriam: This suggests that one of the clients has not implemented a recurring liviliness check into their routing table management.  This suggests that we want an even more granular "Recursive Find Nodes" which gives detailed sourcing information for each node that is checked in the census, so that we know exactly where we got which ENR so that when there is a node in the census giving us bad nodes, we want to be able to point the finger at those nodes.... and identify what client it is.

KolbyML: This PR should fix that https://github.com/ethereum/trin/pull/1036

Kayden told me he is working on improving the enr details page.

I will continue this conversation https://github.com/ethereum/glados/pull/198 so we aren't talking between two PR's

PR #202, opened 2023-11-29
Decoded Enr Viewer

KaydenML:
After image
image

Before Image
image

### DIFF
--- a/glados-web/templates/enr_detail.html
+++ b/glados-web/templates/enr_detail.html
@@ -3,26 +3,96 @@
 {% block title %}ENR Detail{% endblock %}
 
 {% block content %}
-<div class="row">
-    <div class="col">
-        <h1>ENR Details: {{ node.node_id_as_hex() }}</h1>
+<script src="/static/js/trace/enr.js"></script>
+
+<ul>
+    <div class="row">
+        <div class="col">
+            <h1>ENR Details: {{ node.node_id_as_hex() }}</h1>
+            <br/>
+        </div>
     </div>
-</div>
-<div class="row">
-    <div class="col">
-        <ul>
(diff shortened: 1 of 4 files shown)


glados-web/templates/enr_detail.html
```diff
@@ -3,26 +3,96 @@
...
+        if (key === "id") {
+            return value.toString()
+        } else if (key === "secp256k1") {
+            return toHexString(value)
+        } else if (["ip", "tcp", "udp", "ip6", "tcp6", "udp6"].includes(key)) {
+            const field = enr[key]
+            if (!field) {
+                throw new Error("Field not found.")
+            }
+            return field
```
pipermerriam, line 70: This *whole* script tag concerns me with respect to code maintainablity...

I really like that it is broken down into utility functions...  but I also am concerned with the maintenance and verifiable correctness of it...

The functions like `toHexString` and `hexFormatValue` seem like things that we *could*  choose to get from an existing library that is well tested...  Maybe there is even front-end ENR library out there already that we could use?

In it's current form I'm a little concerned with the long term maintenance cost of this code...

pipermerriam (approved):
The javascript in this gives me some long term maintenance concerns, but I am of the opinion that we should...

- get this merged *roughly* as-is
- open an issue noting that we have a longer term maintenance issue that needs to be eventually addressed
- deal with this later

The ideal solution for us to aim for is something that allows us to write tests for any custom javascript that we are doing so that it can be validated as working correctly in an automated manner.  @KaydenML I encourage you to pursue/experiment with how we might do this.

KaydenML:
@pipermerriam I have created an issue regarding the long term maintenance concerns.

PR #210, opened 2023-12-04
fix: prevents oldest unaudited strategy from picking up latest content

mrferris:


carver (commented):
Looks good! Sorry I was slow on the review.

glados-audit/src/selection.rs
```diff
@@ -325,12 +328,17 @@ mod tests {
             let block_hash = [num; 32];
             let content_key =
                 HistoryContentKey::BlockHeaderWithProof(BlockHeaderKey { block_hash });
-            // content table
+            // Content table test data initialization
+            // Check whether row should be new or old data
+            let available_at = match (2..=15).contains(&num) {
+                true => Utc::now() - chrono::Duration::days(1),
```
carver, line 334: Maybe some risk that this will be a little flaky, since it's exactly the (exclusive) cutoff for the audit strategy.

PR #213, opened 2023-12-07
Avoid Infura lock-in for docker-compose and add env to readme

kdeme:
This allows the user to use any node or service as long as it supports the EL JSON-RPC API.


kdeme:
Not sure if the public glados instance uses this docker-compose, but if it does, this is a breaking change.

mrferris (approved):

Issue #205, opened 2023-12-01
glados-web unavailable due to panicks


kdeme:
Getting this error on glados-web when running it manually in combination with glados-audit, glados-monitor and glados-cartographer. Tried both with existing sqlite db and with a fresh sqlite db.

```
[2023-12-01T12:48:10Z DEBUG hyper::proto::h1::io] parsed 12 headers
[2023-12-01T12:48:10Z DEBUG hyper::proto::h1::conn] incoming body is empty
thread 'tokio-runtime-worker' panicked at 'called `Result::unwrap()` on an `Err` value: Type("A null value was encountered while decoding \"id\"")', /home/deme/repos/glados/glados-web/src/routes.rs:206:10
[2023-12-01T12:48:10Z INFO  sqlx::query] SELECT MAX("id") AS "id" …; rows affected: 1, rows returned: 1, elapsed: 34.105µs
    
    SELECT
      MAX("id") AS "id"
    FROM
      "census"
```    

The culpit commit appears to be 5702b8a503cb8f52a14c7419e5559170123f3750, before this commit it still works.

kdeme:
According to @KolbyML solution here is to use Postgres instead of sqlite. Sqlite will be removed and basically is already not supported due to this issue.

kdeme:
@KolbyML I think this issue also exists for postgres deployments. I've deleted my postgres db when I was testing with the docker compose solution and suddenly this error hits also.

KolbyML:
Oh ok I will make a PR to fix this 👍

PR #214, opened 2023-12-07
Add support for Fluffy in the docker compose file

kdeme:
I went the route of using an `GLADOS_PORTAL_CLIENT` env var to select the client, as there are already several of those required anyhow. That way we can stick with just the default `docker-compose.yml`.

I used `extends` instead of `include` as the latter is only for fairly recent docker-compose versions.

Another option would be to create a docker compose file for each client and extends/include some common compose file. Not sure what is more idiomatic in Docker land.

~~This PR is in Draft because I haven't been able to fully test the final version as I'm hitting https://github.com/ethereum/glados/issues/205~~ Have been able to test it in the meanwhile.

### DIFF
--- a/DOCKER_GUIDE.md
+++ b/DOCKER_GUIDE.md
@@ -15,6 +15,9 @@ where Postgres should store its data on the host machine.
 
 - `GLADOS_PROVIDER_URL` to the execution API provider URL.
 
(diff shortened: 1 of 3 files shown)


KolbyML (approved):
Looks good

Issue #200, opened 2023-11-28
Generate graphs from census data.


pipermerriam:
### What is wrong

The new "census" view of the network gives a very detail oriented view of the network.  We need a more visual view

### How can it be fixed

The census is currently running at 15 minute granularity.  There is currently very little change between each snapshot.  We want a way to visualize how the network changes across a number of snapshots.  I'm thinking something like this chart, but where each section represents node-ids in the network, and the flow between them shows nodes joining/leaving the network as the census data changes over time..

![PPC-Visitor-Flow-2480033061](https://github.com/ethereum/glados/assets/824194/6ab756b8-5ef4-4886-aefe-edefb2ed28f3)

- X axis on the chart marks the flow of time, left going into the past, and right going into the future.
- Y axis is determined by node-id.

> I tried to describe it in words, but I need to draw a picture... will update later.

PR #220, opened 2023-12-17
feat: add census nodes timeline chart

mrferris:
**What was the problem?**

Census data was only visible at specific moments of time, rather than over any period of time. 

**How was it fixed?**

The census explorer page shows all nodes seen in the last 24 hours and their census status at each census over that period. 

Nodes are sorted by client string, and then by node ID. Unhighlighted Node IDs on the left axis are meant to be more searchable (Cmd F) than readable. 

Screen Shot 2024-03-11 at 5 56 09 PM

- [x] Efficient scrolling
- [x] Pagination

Subsequent PRs will:
- show errors logged during census
- label trin testnet nodes (eg trin-ams-10)

Implements #200

### DIFF
--- a/entity/src/census_node.rs
+++ b/entity/src/census_node.rs
@@ -26,6 +26,14 @@ pub enum Relation {
         on_delete = "Cascade"
     )]
     Census,
+    #[sea_orm(
+        belongs_to = "super::record::Entity",
+        from = "Column::RecordId"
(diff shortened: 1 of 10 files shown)


KolbyML (approved):
Nothing stands off as wrong, so lets merge it see what happens, worst comes to worse we have to do a small fix in the future. 

:shipit:

PR #229, opened 2024-02-01
fix: remove dependence on libssl

KolbyML:


ogenev (approved):
⛵

KolbyML:
What is the fix ``reqwest = { version = "0.11.6", default-features = false, features = ["rustls-tls"] }`` this is the fix it removes the need to use libssl3.

The reason why our images used ubuntu 20 is because 22 didn't provide libssl3. Now that we don't depend on libssl3 we can update to version 22.

mrferris:
Do we not know why libssl3 is suddenly unavailable on the same ubuntu image? Guessing rust version but I don't see how that would change availability of a lib

KolbyML:
We don't want to depend on libssl3 regardless as the library version isn't supported by default on new OS's. Personally I don't know why it became unavailable probably because of what you mentioned.

KolbyML:
But yeah with the update to the rust container. It probably contained a different version of libssl which causes some kind of dynamic linker conflict when ran

PR #231, opened 2024-02-21
feat: adds fourfours strategy to glados-audit

mrferris:
**What's the problem?**
The approach to 4444s stats taken in #230 requires pre-loading the DB with all 45mil+ content keys. This is quite an operational overhead. 

**What's the solution?**
This PR adds a dedicated `FourFours` audit strategy that has access to a `Web3` provider that it uses to randomly select  hashes for blocks between genesis and merge, stores their content keys in the DB, and then audits them. 

It also adds 4444s statistics to the front page, and the filtering option to the Audit Dashboard.

### DIFF
--- a/glados-audit/src/lib.rs
+++ b/glados-audit/src/lib.rs
@@ -37,6 +37,8 @@ pub(crate) mod validation;
 pub struct AuditConfig {
     /// For Glados-related data.
     pub database_url: String,
+    /// For getting on-the-fly block information.
+    pub provider_url: String,
     /// Specific strategies to run.
     pub strategies: Vec<SelectionStrategy>,
     /// Weight for each strategy.
@@ -65,6 +67,7 @@ impl AuditConfig {
                 "Selected concurrency set."
             )
         }
+
         let strategies = match args.strategy {
             Some(s) => s,
             None => {
@@ -73,6 +76,7 @@ impl AuditConfig {
                     SelectionStrategy::Random,
                     SelectionStrategy::Failed,
                     SelectionStrategy::SelectOldestUnaudited,
+                    SelectionStrategy::FourFours,
                 ]
             }
         };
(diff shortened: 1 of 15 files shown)


glados-audit/src/lib.rs
```diff
@@ -83,6 +89,14 @@ impl AuditConfig {
                 SelectionStrategy::Random => args.random_strategy_weight,
                 SelectionStrategy::Failed => args.failed_strategy_weight,
                 SelectionStrategy::SelectOldestUnaudited => args.oldest_strategy_weight,
+                SelectionStrategy::FourFours => {
+                    if args.provider_url.is_empty() {
+                        warn!("No provider URL provided, disabling fourfours audit strategy.");
```
pipermerriam, line 94: These seems like surprising implicit behavior.  Seems like if the 4444's strategy is enabled and no provider is present it should be an error?

glados-audit/src/selection.rs
```diff
@@ -95,6 +103,40 @@ async fn select_latest_content_for_audit(
...
+async fn select_fourfours_content_for_audit(
+    tx: mpsc::Sender<AuditTask>,
+    conn: DatabaseConnection,
+    w3: web3::Web3<web3::transports::Http>,
+) -> ! {
+    let mut interval = interval(Duration::from_secs(5));
+
+    loop {
+        interval.tick().await;
+        let block_number = thread_rng().gen_range(1..15537393);
```
pipermerriam, line 118: This number should probably be defined as a constant somewhere **AND** maybe it should be CLI configurable/overideable?

glados-audit/src/selection.rs
```diff
@@ -95,6 +103,40 @@ async fn select_latest_content_for_audit(
...
+            .await
+            .unwrap()
+            .unwrap()
+            .hash
+            .unwrap();
+
+        let items_to_audit =
+            store_block_keys(block_number, block_hash.as_fixed_bytes(), &conn).await;
+        debug!(
+            strategy = "fourfours",
```
pipermerriam, line 132: *nitpick*  Any reason not to do `"4444"` which I think is more visually distinguishable...

glados-audit/src/lib.rs
mrferris: > surprising implicit behavior
had the same thought, forgot to delete, thanks 👍

KolbyML (approved):
:shipit: looks cool

PR #233, opened 2024-03-01
feat: add time on hoverover to stats chart

mrferris:
@njgheorghita:
> it would be nice if on the success rate graph, while hovering over the graph if it could display the time at which the cursor is that would be very helpful

Agreed 👍


njgheorghita (approved):
:rocket:

Just a few more thoughts on the glados landing page... but feel free to ignore it b/c it's not high-priority imo
1. it would be nice to have the selector for the graph also change the stats displayed in the main chart
2. maybe an option to see a smoothed version / trend line in the graph? like a 1 day moving average or something like that

PR #241, opened 2024-03-18
chore: use latest ethportal-api via git import

mrferris:
**What was wrong?**
#113 should have been merged earlier (thank you @perama-v!) and its necessary rebase has grown unruly, so I'm doing it in this PR. 

This PR updates ethportal-api to use the latest from github (crates.io cannot be updated due to reth's git dependency), as well as some necessary dependency updates, refactoring and dependency-issue fixing:

- The `reth-ipc` import began showing a cargo error that said only the latest version (`0.2.0-beta.3`) from paradigm's github was available, and that the version we'd been using (`0.1.0-alpha.6`) was no longer available. All of the reth versions seem to [still be available](https://github.com/paradigmxyz/reth/releases/tag/v0.1.0-alpha.6), so I'm not sure what's causing this or why it happened once I switched the `ethportal-api` import. I don't think IPC support is a high priority (cartographer doesn't support it) and am in favor of removing it anyway, so I have removed it here and will either remove mention of support for it or fix this issue in a next PR. 
- `discv5` versions were clashing, turns out glados itself wasn't even making use of the import, which is something interesting to note about `udeps`, which I used to find unused dependencies: it won't count a dependency as unused if another one of your dependencies uses it. Which is fine build-time/binary-bloat wise, but not Cargo.toml-bloat wise.


KolbyML (approved):
:shipit:

PR #244, opened 2024-03-19
fix: inaccurate content key timestamps

### LINKED ISSUE #238: FourFours data showing invalid "First Available At" value
### What is wrong

There is FourFours data showing a "First Available At" that is far too recent to be valid.

Screenshot from 2024-03-12 10-59-10

### How should it be fixed

:shrug: 

Not sure, but it seems that anything that is FourFours related should be showing a data that is more than a year ago.

mrferris:
Fixes #238 

Also fixes issue where `glados-monitor` was using the time that the content key enters the DB, which could be many seconds after the block hit the chain.

Screen Shot 2024-03-19 at 4 19 01 PM

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2333,6 +2333,7 @@ name = "glados-monitor"
 version = "0.1.0"
 dependencies = [
  "anyhow",
+ "chrono",
  "clap 4.5.3",
  "entity",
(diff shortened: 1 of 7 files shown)


KolbyML (approved):
:shipit:

PR #245, opened 2024-03-20
feat: adds node nicknames (aka trin testnet names) to census explorer & node details page

mrferris:
**What was wrong?**

Couldn't tell on census explorer which nodes are trin testnet nodes.

**How was it fixed?**

The `node` model was given a `HashMap`, mapping node IDs to nicknames. 

The census explorer sorts its rows to have all trin testnet nodes in the same block, under the fluffy nodes. 

And then as a little nice-to-have, I added to the node page a trin testnet node's name and links to its grafana metrics. Maybe we don't keep that long term, but I think it will be convenient for debugging.
Screen Shot 2024-03-20 at 7 58 34 PM


KolbyML (approved):
:shipit:

njgheorghita (approved):

PR #246, opened 2024-03-23
fix: improves sql efficiency

mrferris:
**What was wrong?**

Some of our sql operations are inefficient and are taking seconds to run at scale while redlining cpu/mem on the glados host.

**What was fixed?**

1. `content` table has an index on `(protocol-id, content-key)`, but the `get_or_create` function was only doing a filter on `content-key` so this index wasn't being used and `content` creation was taking a long time. Filtering on `protocol-id` makes both indexes get used.
2. `content_audit` table was missing an index for the foreign key to the `content` table. Joining the `content_audit` table on the `content` table is necessary for the `latest` and `oldest` strategies, and those queries were taking a long time.
3. The `random` selection strategy had two inefficiencies: 1.) we were using `COUNT` to get all of the content IDs so that we could randomly sample from them, this requires scanning the entire table which was taking seconds. Using `MAX` instead makes it instant. 2.) with the IDs for the content in the DB that we wanted to lookup, we were doing a separate sql query for each one. Doing a single sql query to get all (up to 100) keys results in a 10x speed-up.

### DIFF
--- a/migration/src/lib.rs
+++ b/migration/src/lib.rs
@@ -10,6 +10,7 @@ mod m20230511_104838_create_execution_metadata;
 mod m20230511_104937_create_key_value;
 mod m20231107_004843_create_audit_stats;
 mod m20240213_190221_add_fourfours_stats;
+mod m20240322_205213_add_content_audit_index;
 
 pub struct Migrator;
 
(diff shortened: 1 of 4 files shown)


migration/src/lib.rs
```diff
@@ -10,23 +10,25 @@ mod m20230511_104838_create_execution_metadata;
...
 mod m20240213_190221_add_fourfours_stats;
+mod m20240322_205213_add_content_audit_index;
 
 pub struct Migrator;
 
 #[async_trait::async_trait]
 impl MigratorTrait for Migrator {
     fn migrations() -> Vec<Box<dyn MigrationTrait>> {
         vec![
+            Box::new(m20230508_111707_create_census_tables::Migration),
```
pipermerriam, line 21: Why was this added?  Was it missing?

pipermerriam (approved):
Good finds here.

migration/src/lib.rs
mrferris: No, looks like `cargo fmt` auto-sorted these entries by the prefixed dates. It was removed a few lines below this.

PR #249, opened 2024-03-26
feat: audit stats history pagination by week

mrferris:
adds ability to scroll back week-by-week on glados success stats


KolbyML (approved):
Looks good I was going to comment on inputting negative values, but it looks like it shouldn't cause any issues if that happens so :shipit:

PR #251, opened 2024-04-01
feat: use postgres instead of sqlite for tests & ci

mrferris:
**What's the problem?**

Our production deployments run on postgres, but all our tests run on sqlite. These two are different enough to make our tests (and thus CI results) not necessarily reflective of the behavior of our production deployment. 

**What's the solution?**

Using the `pgtemp` crate to create a postgres instance for unit tests to use. It uses a local installation of postgres instead of installing docker and then downloading postgres, which would make CI take much longer. 

This PR also removes all mentions of sqlite from the repo, and adds instructions for quickly spinning up a local postgres instance via docker.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -85,6 +85,9 @@ jobs:
                 name: Install libssl-dev for openssl-sys
                 command: sudo apt install -y libssl-dev clang
             - run:
+                name: Install postgres for testing
+                command: sudo apt install -y postgresql postgresql-client
+            - run:
                 name: Install rustfmt
                 command: rustup component add rustfmt
             - run:
@@ -101,7 +104,7 @@ jobs:
                 command: cargo build --workspace
             - run:
                 name: Test Glados workspace
-                command: cargo test --workspace -- --nocapture
+                command: export POSTGRES_PATH=$(find /usr/lib/postgresql/ -type d -name "bin") && PATH=$PATH:$POSTGRES_PATH cargo test --workspace -- --nocapture
 workflows:
   merge-test:
     jobs:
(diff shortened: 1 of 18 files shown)


KolbyML (approved):
:shipit:

.circleci/config.yml
```diff
@@ -95,13 +98,13 @@ jobs:
                 command: rustup component add clippy
             - run:
                 name: Run Clippy
-                command: cargo clippy --all --all-targets --all-features --no-deps -- --deny warnings
+                command:  cargo clippy --all --all-targets --all-features --no-deps -- --deny warnings
```
pipermerriam, line 101: *nit* bonus/extra space.

.circleci/config.yml
```diff
@@ -95,13 +98,13 @@ jobs:
...
             - run:
                 name: Build Glados workspace
                 command: cargo build --workspace
             - run:
                 name: Test Glados workspace
-                command: cargo test --workspace -- --nocapture
+                command: PATH=$PATH:$(find /usr/lib/postgresql/ -type d -name "bin") cargo test --workspace -- --nocapture
```
pipermerriam, line 107: Maybe this deserves a comment?  Does it not land in a static location in CI?

glados-audit/src/cli.rs
```diff
@@ -1,7 +1,7 @@
 use clap::{ArgAction, Parser, Subcommand, ValueEnum};
 use entity::content_audit::SelectionStrategy;
 
-const DEFAULT_DB_URL: &str = "sqlite::memory:";
+const DEFAULT_DB_URL: &str = "postgres://postgres:password@localhost:5432/glados";
```
pipermerriam, line 4: Does it still make sense to have a *default* for this parameter now that we're using postgres and it requires that a database be setup separate from running glados?

pipermerriam (approved):

Issue #235, opened 2024-03-04
Initial state-root based auditing strategy for state network.


pipermerriam:
### What is wrong

We need an auditing strategy for state network.

Ideally, the strategy avoids some of the inherent complexity such as:

- lack of easy access to knowing what state data exists
- sheer quantity of total state data that would need to be loaded
- complexity of state data existing across many state roots

### How can it be done.

I propose an initial strategy that is based on state roots.

The only initial data that would need to be loaded into the database would be state roots.

The strategy would perform random walks of the state data beginning at the state root and terminating when either data is determined to be unavailable or the walk terminates  in a leaf node.

The walk begins with a state root, which can be used to retrieve the trie node associated with the state root.  At each stage the following logic would be applied.

- If the node is a leaf node **and** the account is an EOA: Terminate the walk and mark it as a success.
- If the node is a leaf node **and** the account is a contract:  Fetch the contract storage trie state root node and continue.
- If the node is a branch node: Pick a random child node hash, fetch that node and continue
- If the node is an extension: Follow it to it's child and fetch the child node.

This strategy will have some shortcomings, but it should be simple to implement and it will be a viable jumping off point for more detailed and thorough auditing strategies.

There are a few choices that should be considered:

- Upon encountering trie nodes, should we save them to the database?  If we do, then we can then perform individual audits of that data directly.
- Upon failure: what data should we save?
- Upon success: what data should we save?

morph-dev:
> If the node is a leaf node and the account is a contract: Fetch the contract storage trie state root node and continue.

Just realized that this doesn't work.

In order to iterate contract storage trie, we have to know the address of the smart contract (in order to construct the content key). However, if we discovered smart contract account in the account trie using steps from above, we don't know its address. We know its path from the root, but path is actually `keccak(address)`, and there is no way to reverse it.

To work around it, we would have to build a path->address map for smart contracts.

Alternatively, we can decide on a smart contract address (e.g. picking random transaction that was smart contract call), iterate account trie to find its storage trie root, and randomly walk the storage trie as already described.

pipermerriam:
Simplest solution to this is to store the pre-image map in the network which I think is a reasonable way to do this, except that where to source the pre-images is the difficult part.  I think that means that our bridges need to produce this mapping as part of their data export since I think the only reliable way to get 100% coverage on the pre-images would be to source them from tracing the EVM execution?

cc @KolbyML

morph-dev:
[...]

PR #259, opened 2024-05-10
feat: implement "Initial state-root based auditing strategy for state network"

KolbyML:
Implements https://github.com/ethereum/glados/issues/235

This PR adds
- distinguishing audit strategy types by networks (not all networks will support the same audit strategies)
- And then implements that issue by doing random walks of the trie at randomly selected blocks

I set the tree random walker to do 10 nodes a second, maybe this is too low. None of the RFC require uTP to my knowledge as the intermediate trie nodes are just that small


glados-audit/src/state.rs
```diff
@@ -0,0 +1,260 @@
...
+async fn save_state_audit(
+    conn: &DatabaseConnection,
+    content_key: StateContentKey,
+    audit_result: bool,
+    block_number: i32,
+    client: PortalClient,
+) -> Result<()> {
+    let content_model = store_content_key(
+        &content_key,
+        "state_itermidate_trie_node",
```
mrferris, line 40:
```suggestion
        "state_intermediate_trie_node",
```

KolbyML:
@mrferris I fixed the spelling issue, is there any other concerns with this PR?

mrferris (changes requested):

KolbyML:
@mrferris I believe I resolved the PR concerns, let me know if you have any other concerns

glados-audit/src/state.rs
```diff
@@ -0,0 +1,303 @@
...
+
+async fn process_trie_node(
+    conn: &DatabaseConnection,
+    portal_client: PortalClient,
+    block_number: i32,
+    path: &[u8],
+    content_key: StateContentKey,
+    trie_node: Node,
+    available_at: DateTime<Utc>,
+) -> Result<Option<(StateContentKey, usize)>, (anyhow::Error, StateContentKey)> {
```
morph-dev, line 83: What is `usize` in the return type supposed to be?
Why is it `0` for extension node and `index` for branch node? This doesn't seem correct.

glados-audit/src/state.rs
```diff
@@ -0,0 +1,303 @@
...
+
+async fn random_state_walk(
+    conn: &DatabaseConnection,
+    state_root: B256,
+    portal_client: PortalClient,
+    client: HttpClient,
+    block_number: i32,
+    available_at: DateTime<Utc>,
+) -> Result<(), (anyhow::Error, StateContentKey)> {
+    let mut stack = vec![StateContentKey::AccountTrieNode(AccountTrieNodeKey {
```
morph-dev, line 172: Why is this stack? We always have at most one element in, right?

morph-dev (changes requested):
Since I'm not familiar with glados codebase, I focused mostly on code in `glados-audit/src/state.rs`.

glados-audit/src/lib.rs
```diff
@@ -156,33 +181,71 @@ pub async fn run_glados_command(conn: DatabaseConnection, command: cli::Command)
...
+            conn.clone(),
         ));
     }
-    // Collation of generated tasks, taken proportional to weights.
-    let (collation_tx, collation_rx) = mpsc::channel::<AuditTask>(100);
-    tokio::spawn(start_collation(collation_tx, task_channels));
-    // Perform collated audit tasks.
-    tokio::spawn(perform_content_audits(config, collation_rx, conn));
+
+    if config.history {
```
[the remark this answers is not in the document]
KolbyML: I think that should wait till a future PR, till we know what we want to do with beacon

entity/src/state_roots.rs
```diff
@@ -0,0 +1,63 @@
...
+use alloy_primitives::B256;
+use anyhow::Result;
+use chrono::{DateTime, Utc};
+use rand::Rng;
+use sea_orm::{entity::prelude::*, QuerySelect, Set};
+
+#[derive(Clone, Debug, PartialEq, DeriveEntityModel, Eq)]
+#[sea_orm(table_name = "state_roots")]
+pub struct Model {
+    #[sea_orm(primary_key)]
```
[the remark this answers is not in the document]
KolbyML: sure I don't think it will make a difference but might as well

glados-audit/src/state.rs
```diff
@@ -0,0 +1,303 @@
...
+                    Ok(Some((
+                        StateContentKey::AccountTrieNode(AccountTrieNodeKey {
+                            path: Nibbles::try_from_unpacked_nibbles(path).expect("Bug building path in random_state_walk"),
+                            node_hash: hash_node.hash,
+                        }),
+                        0,
+                    )))
+                }
+                other_node => {
+                    Err((anyhow!(
```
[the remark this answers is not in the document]
KolbyML: I am aware of this, but since I am not dealing with the contract storage trie, I decided that could be done in a future PR, when we walk the contract storage trie

glados-audit/src/state.rs
```diff
@@ -0,0 +1,303 @@
...
+            }
+        }
+        Node::Extension(extension_node) => {
+            let hash_node =
+                &extension_node.read().expect("Read should never fail").node;
+            match hash_node {
+                Node::Hash(hash_node) => {
+                    Ok(Some((
+                        StateContentKey::AccountTrieNode(AccountTrieNodeKey {
+                            path: Nibbles::try_from_unpacked_nibbles(path).expect("Bug building path in random_state_walk"),
```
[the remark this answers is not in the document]
KolbyML: good catch

glados-audit/src/state.rs
KolbyML: Because I was originally writing this to walk a whole tree and DFS uses a stack, I thought I would leave it as it doesn't break anything, but it can also be useful for the future

glados-audit/src/state.rs
KolbyML: I updated the return of extension node. I am returning the next bit of the path to continue the walk

KolbyML:
@morph-dev let know if you have any other concerns, ready for another review

glados-audit/src/lib.rs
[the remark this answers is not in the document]
morph-dev: My philosophy is that *"if you touch the code, you should leave it in a better state than you find it"*. This applies to code readability, code design, and test. This greatly helps with technical debt, but it does sacrifice development speed (at least for the short term).

This is just my opinion and I don't want to enforce it. So it's fine if you just acknowledge it.

glados-audit/src/state.rs
```diff
@@ -0,0 +1,309 @@
...
+    let mut path: Vec<u8> = vec![];
+    while let Some(content_key) = stack.pop() {
+        match StateNetworkApiClient::recursive_find_content(&client, content_key.clone()).await {
+            Ok(response) => match response {
+                ContentInfo::Content {
+                    content: content_value,
+                    ..
+                } => match content_value {
+                    StateContentValue::TrieNode(encoded_trie_node) => {
+                        let trie_node = encoded_trie_node.node.as_trie_node().expect("Trie node received from the portal network should be decoded as a trie node");
```
morph-dev, line 192: I'm not sure if this is normally done, but should we check whether retrieved value (in this case trie node) actually matches the content_key?
In this case, I would say it's enough to check that `EncodedTrieNode.node_hash` matches the `node_hash` from content key.

glados-audit/src/state.rs
```diff
@@ -0,0 +1,309 @@
...
+                    content_key,
+                ));
+            }
+        }
+        // Limit check for new tasks to 10/sec
+        sleep(Duration::from_millis(100)).await;
+    }
+    Err((
+        anyhow!("Walk exhausted without finding a leaf node or failing, maybe there is a bug? this should not happen."),
+        StateContentKey::AccountTrieNode(AccountTrieNodeKey {
```
morph-dev, line 238: Is this supposed to be the same content key as the one created at line 178? If so, maybe save it in a variable: `root_content_key`?

glados-audit/src/state.rs
```diff
@@ -0,0 +1,309 @@
...
+            };
+
+            let transport = PortalApi::parse_client_url(portal_client.api.client_url.clone())
+                .await
+                .expect("Failed to parse client url.");
+            let client = match transport {
+                Transport::HTTP(http) => http.client,
+            };
+
+            if let Err((err, content_key)) = random_state_walk(
```
morph-dev, line 283: I feel like `save_state_audit` should be called here, both in case of error (as it is now) and for success (not inside `process_trie_node`).

morph-dev (changes requested):

glados-audit/src/state.rs
```diff
@@ -0,0 +1,260 @@
...
+        node_hash: state_root,
+    })];
+    let mut path: Vec<u8> = vec![];
+    while let Some(content_key) = stack.pop() {
+        match StateNetworkApiClient::recursive_find_content(&client, content_key.clone()).await {
+            Ok(response) => match response {
+                ContentInfo::Content {
+                    content: content_value,
+                    ..
+                } => match content_value {
```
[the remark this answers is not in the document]
KolbyML: Very cool I didn't think of doing it like this

glados-audit/src/state.rs
KolbyML: Fine it isn't a big deal as if we need it we can always change it back

glados-audit/src/lib.rs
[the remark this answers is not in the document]
KolbyML: I am not sure the proper way to acknowledge this but I did read it.

I would want to wait till the next PR for beacon audit strategies to be done for a refactor to be done. As combining them creates more work and complexity to handle for whenever a beacon audit refactor happens,

I implemented what was suggested cause it isn't the biggest concern on earth, but I do think this is less clean and more complex. I don't think less code duplication is worth that cost in this example

KolbyML:
@morph-dev ready for another round, I believe I resolved all of the concerns but as always if I missed any please let me know.

glados-audit/src/state.rs
```diff
@@ -73,20 +73,12 @@ async fn save_state_audit(
...
-    conn: &DatabaseConnection,
-    portal_client: PortalClient,
-    block_number: i32,
-    path: &[u8],
-    content_key: StateContentKey,
+    content_key: AccountTrieNodeKey,
     trie_node: Node,
-    available_at: DateTime<Utc>,
-) -> Result<Option<(StateContentKey, usize)>, (anyhow::Error, StateContentKey)> {
+) -> Result<(AccountTrieNodeKey, bool), (anyhow::Error, AccountTrieNodeKey)> {
```
morph-dev, line 78: nit: This feels like an abuse of `Result` type. Knowing you are going to keep refactoring this as we add support for storage trie, I would say it's good for now. But return type could be something like this:
```rust
enum ProcessAccountTrieNodeResult {
  LeafReached,
  NextNode(AccountTrieNodeKey),
  Error(String), // or if you really want Error(anyhow::Error)
}
```

Maybe also pass `EncodedTrieNode` (instead of `Node`) and do other two checks here as well.

morph-dev (changes requested):
Almost there. One important comment and two nits.

glados-audit/src/state.rs
```diff
@@ -168,79 +153,76 @@ async fn process_trie_node(
...
             .expect("Bug building path in random_state_walk"),
         node_hash: state_root,
-    })];
-    let mut path: Vec<u8> = vec![];
-    while let Some(content_key) = stack.pop() {
-        match StateNetworkApiClient::recursive_find_content(&client, content_key.clone()).await {
+    };
+    let mut current_content_key = root_content_key.clone();
+    loop {
+        let content_value = match StateNetworkApiClient::recursive_find_content(
```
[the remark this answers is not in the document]
KolbyML: updated, didn't notice it

glados-audit/src/state.rs
KolbyML: > nit: This feels like an abuse of `Result` type. Knowing you are going to keep refactoring this as we add support for storage trie, I would say it's good for now. But return type could be something like this:
> 
> ```rust
> enum ProcessAccountTrieNodeResult {
>   LeafReached,
>   NextNode(AccountTrieNodeKey),
>   Error(String), // or if you really want Error(anyhow::Error)
> }
> ```


I think it is a good idea, I would be happy to do this in a future PR when we refactor this for the storage trie like you said

morph-dev (approved):
LGTM, but wait for approval from @mrferris as well (since I'm not familiar with glados codebase and I mostly focused on the state related code).

mrferris (approved):

PR #262, opened 2024-05-10
doc: fix postgres launch command in readme

carver:
It was missing the environment variable flag, so gave this error in the shell:

```
docker: invalid reference format: repository name must be lowercase.
```

(trying to interpret POSTGRES_PASSWORD=password as the repository)


mrferris (approved):
Good catch

PR #263, opened 2024-05-13
refactor: update to use alloy-primitives so we can update ethportal-api

KolbyML:
ethportal-api switched to alloy-primitives, updating glados to use it as well so we can update the version of ethportal-api used


pipermerriam (approved):
Looks straight forward :shipit:

PR #264, opened 2024-05-15
feat: add nodes that will accept count on contentkey_detail page

KolbyML:
image


KolbyML:
I tested it and it works

mrferris (approved):
Also tested👍

PR #265, opened 2024-05-18
Add data coverage chart

### LINKED ISSUE #261: Generate a chart of the "best case" network coverage, based on nodes' claimed radii
Stacked area chart, showing network replication (based on the claimed radius of the node)
```[tasklist]
### Tasks
- [x] Paper protototype of how xor distance plays out in a replication chart
- [x] Generate a stacked area chart in d3 (or maybe just a bar chart), using dummy data
- [x] Calculate chart data, based on node ID & radius data
- [x] Test tradeoffs between high-resolution/low-speed and low-resolution/high-speed display
- [x] Clean up for merge
- [x] Validate against production data
- [x] Generate keycount for every node, for every prefix
```

carver:
Fix #261 

Adds a chart to the home page, showing how much coverage there is in the best case.

Here's an example from a real census run on my machine a few days ago:
image

The basic approach is to split the keyspace into 256 1-byte buckets. Then show how many nodes fully cover that keyspace, and how much partial coverage there is of those buckets.

This mostly reuses the existing data from the node radius chart, plus a bit more precision in the radius calculation sent by the server.

Also, I added a tooltip to help you see which bucket is showing which data, and the precise value of coverage. I am a little unsatisfied with the method of identifying the bucket prefix (back-calculating from the x-value), but I couldn't find a better solution after digging a bit. I'm open to suggestions.

Note: there are a few little fix-up commits working on the logs and readme. It may be easier to review by commit.

TODO:
- [x] remove TODO comments, they should be fully captured in #267 now
- [ ] squash comment removal and tooltip into the "add chart" commit

### DIFF
--- a/glados-web/assets/js/radiusdensity.js
+++ b/glados-web/assets/js/radiusdensity.js
@@ -1,3 +1,156 @@
+function radius_stacked_chart(data) {
+    const bucket_bit_width = 8;
+    const num_buckets = 2 ** bucket_bit_width;
+    const margin = {top: 10, right: 15, bottom: 60, left: 65},
+        width = 1060 - margin.left - margin.right,
+        height = 425 - margin.top - margin.bottom;
+
+    // append the svg object to the body of the page
+    const svg = d3.select("#census-stacked")
+        .append("svg")
+        .attr("width", width + margin.left + margin.right)
+        .attr("height", height + margin.top + margin.bottom)
+        .append("g")
+        .attr("transform",
+            `translate(${margin.left}, ${margin.top})`);
+
+    // Add X axis
+    const x = d3.scaleLinear()
+        .domain([0, num_buckets])
+        .range([ 0, width ]);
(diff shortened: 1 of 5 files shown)


glados-web/assets/js/radiusdensity.js
```diff
@@ -1,3 +1,162 @@
...
+    svg.append("text")
+        .attr("class", "x label")
+        .attr("text-anchor", "middle")
+        .attr("x", width / 2)
+        .attr("y", height + margin.top + margin.bottom - 15)
+        .text("First byte of Content ID");
+
+    // Add Y axis
+    const y = d3.scaleLinear()
+        .domain([0, 16])
```
mrferris, line 45: This should probably scale upwards when the number of nodes covering surpasses 16, right?

mrferris (approved):
Looks good, tested on my machine

PR #272, opened 2024-05-23
Data coverage chart v1.1

carver:
- Dynamic y-axis size (sorry I missed that on the merge of #261 )
- Ignore large nodes

### DIFF
--- a/glados-web/assets/js/radiusdensity.js
+++ b/glados-web/assets/js/radiusdensity.js
@@ -1,4 +1,6 @@
 function radius_stacked_chart(data) {
+    // If a node has a larger radius fraction than this, ignore it in the coverage calculation
+    const maxRadiusFraction = 0.1;
     const bucket_bit_width = 8;
     const num_buckets = 2 ** bucket_bit_width;
     const margin = {top: 10, right: 15, bottom: 60, left: 65},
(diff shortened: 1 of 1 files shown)


glados-web/assets/js/radiusdensity.js
```diff
@@ -14,6 +16,44 @@ function radius_stacked_chart(data) {
...
+    // Indexed data
+    // based on shape of data produced in this example: https://d3js.org/d3-shape/stack#_stack
+    const full = Array(num_buckets).fill(0);
+    const part = Array(num_buckets).fill(0.0);
+    data.forEach(function (node, idx, arr) {
+      const nodePrefix = Number(BigInt(node.node_id) >> (64n - BigInt(bucket_bit_width)));
+      const radiusPrefix = node.radius_top;
+      if (radiusPrefix / 256 > maxRadiusFraction) {
+        // Ignore nodes with large radius. They are probably new nodes that are syncing
+        return ;
```
mrferris, line 28: One thought here is that I'm not sure whether this is throwing out the node entirely, or just having it count toward this 10% segment that we assume it covers at the maximum. I don't think it should throw it out entirely. But I don't feel too strongly, maybe we do just ignore it until it has a reasonable radius.

mrferris (approved):

mrferris:
Small nit, the chart doesn't seem to take up the entire pane on my screen. 
Screenshot 2024-05-24 at 11 41 41 AM

For some reason setting `width = 1060 + margin.left + margin.right,` instead of `width = 1060 - margin.left - margin.right,` fixes it. Not sure exactly why that would fix anything though. 

Screenshot 2024-05-24 at 11 42 59 AM

PR #277, opened 2024-05-29
feat: make api's for graphs we use instead of templating json

KolbyML:
we talked about this during the glados talk

### DIFF
--- a/glados-web/src/lib.rs
+++ b/glados-web/src/lib.rs
@@ -91,6 +91,14 @@ pub async fn run_glados_web(config: Arc<State>) -> Result<()> {
             get(routes::hourly_success_rate),
         )
         .route(
+            "/api/latest-census-node-diversiity-data",
+            get(routes::get_radius_diversity_data),
+        )
+        .route(
+            "/api/latest-census-node-radius-data",
+            get(routes::generate_radius_graph_data),
+        )
+        .route(
             "/api/is-content-in-deadzone/:content_key",
             get(routes::is_content_in_deadzone),
         )
--- a/glados-web/src/routes.rs
+++ b/glados-web/src/routes.rs
@@ -107,7 +107,9 @@ impl Display for ClientDiversityResult {
     }
 }
 
-async fn generate_radius_graph_data(state: &Arc<State>) -> Vec<CalculatedRadiusChartData> {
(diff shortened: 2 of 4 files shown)


mrferris:
I think that the result of the discussion was that we wanted either/both APIs for getting just important values (eg number of nodes seen in the last 24 hours), or fully embeddable HTML elements. I think that returning an entire JSON object to the ethportal website leaves way more front-end work/maintainence on the ethportal website than we'll want.

KolbyML:
> I think that returning an entire JSON object to the ethportal website leaves way more front-end work/maintainence on the ethportal website than we'll want.

I don't think this is how it works. This PR doesn't create any additional maintenance work, it actually creates less. Currently the website is built as if it was 2012. This PR better aligns the project with modern web practices.

> I think that returning an entire JSON object

We are currently doing this already. So I am not sure what is being referred to here. before this PR we are embedding json into a webpage which is extreme malpractice. Doing this causes our html pages unable to be cached by CDN's. The difference here is instead of embedding json (which is very bad), we are fetching json which is proper practice.

I could go further into how this helps productivity, but I believe the issues I brought to light above would have to be addressed before that kind of discussion would be productive.

> fully embeddable HTML elements

This is a completely separate issue and is not directly related..

mrferris:
There's been a misunderstanding regarding the conversation that was had at the summit. 

We were not talking about having glados call JSON endpoints like you're doing here (I agree that loading HTML/JS that itself fetches JSON is better, which is why the stats history and census explorer use that approach, but calling HTML templating "malpractice" is a bit much, and certainly not an issue that would've been worth discussing at summit). 

We were talking about giving ethportal.net (a separate website from glados) access to some of the components within glados for marketing purposes. And having ethportal.net load a big JSON object containing all of the census data and then do its own processing to obtain simple metrics like "how many live nodes on the portal network?" is outside the scope of ethportal.net. It'd be preferable to instead have a simple endpoint that returns a single number that ethportal.net can hit and not have to do any processing on. ethportal.net's processing shouldn't have to change if the data format going into glados's front end changes, they should be decoupled. 

That said, if you add a PR description this can move forward.

KolbyML:
@mrferris sounds good I updated the descriptions

PR #281, opened 2024-05-31
fix: remove unused network explorer and content dashboard

mrferris:
- Remove unused "Content Dashboard" and "Network Explorer" pages
- Publicly rename "Census Explorer" to "Network Explorer" because it's simpler to understand


KolbyML:
I think hide pages doesn't mean remove them, they can be removed from the navigation bar if needed, I believe the extent of this PR is out of the scope we discussed at the summit

mrferris:
I don't think the word "hide" was used, these pages aren't being hidden, updated, and then un-hidden. They're being completely re-written, I can't think of anything on either of them that will be re-used.

KolbyML (approved):
My bad looks good :shipit:

PR #283, opened 2024-06-01
feat: angelfood docker compose file

KolbyML:
We recently set up our first testnet angelfood.

we want to have a glados instance for it using glados.

In the PR I make a new docker compose file configured for running an instance of Glados for out testnet Angelfood. With state audit enabled along with extending our client docker compose file to be able to run other networks

It can be found here https://glados-angelfood.ethdevops.io/audits/

### DIFF
--- a/docker-compose-angelfood.yml
+++ b/docker-compose-angelfood.yml
@@ -1,0 +1,83 @@
+version: '3'
+services:
+
+  glados_postgres:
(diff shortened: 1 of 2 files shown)


pipermerriam (approved):
This is a very rubber-stamp approval.  I don't have much experience with docker compose so there's very little chance I'd catch an error.  :+1:

PR #287, opened 2024-06-25
refactor: updated glados to use ethportal-api instead of custom json-rpc code

KolbyML:
image
image

In glados we were using custom code for interacting with Portal Clients JSON-RPC. To PR updates glados to use our standard JSON-RPC interfaces through our `ethportal-api` crate

### DIFF
--- a/glados-core/src/jsonrpc.rs
+++ b/glados-core/src/jsonrpc.rs
@@ -1,28 +1,20 @@
-use std::str::FromStr;
 use std::{path::PathBuf, time::Duration};
 
-use alloy_primitives::hex::{self, FromHex};
-use alloy_primitives::{B256, U256};
+use anyhow::{anyhow, Ok};
 use entity::content;
-// use ethereum_types::{H256, U256};
-use jsonrpsee::{
-    core::{client::ClientT, params::ArrayParams},
-    http_client::{HttpClient, HttpClientBuilder},
-    rpc_params,
-};
-use serde::{Deserialize, Serialize};
-use serde_json::{
-    json,
-    value::{to_raw_value, RawValue},
-    Value,
-};
-
-use ethportal_api::utils::bytes::{hex_decode, hex_encode, ByteUtilsError};
-use thiserror::Error;
-use tracing::{error, info};
-use url::Url;
-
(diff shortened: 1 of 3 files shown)


glados-core/src/jsonrpc.rs
```diff
@@ -223,212 +60,125 @@ impl PortalClient {
...
+                    Some(Content {
+                        raw: content.encode(),
+                    }),
+                    json!(trace).to_string(),
                 ))
             }
-            Err(err) => match err {
-                JsonRpcError::ContentNotFound { trace } => Ok((None, trace.unwrap_or_default())),
-                _ => Err(err),
-            },
```
mrferris, line 348: This logic was removed, which is making audits run without ever being labelled a success or failure because the error (defined in the Portal RPC spec) propagates and makes it appear that the audit itself errored.

mrferris (changes requested):

glados-core/src/jsonrpc.rs
KolbyML: Ok I added that back if you could give it another look now

glados-core/src/jsonrpc.rs
```diff
@@ -223,212 +143,146 @@ impl PortalClient {
...
-
-                let content_raw = hex_decode(&query_result.content)?;
-                Ok(Some(Content { raw: content_raw }))
+        match content.protocol_id {
+            content::SubProtocol::History => {
+                let result = HistoryNetworkApiClient::recursive_find_content(
+                    &self.client,
+                    content.content_key.clone().try_into()?,
+                )
+                .await?;
```
mrferris, line 181: `get_content_with_trace` looks great, but `get_content` is returning the error rather than checking whether it's a `ContentNotFound` error. Should do the same as `get_content_with_trace`.

mrferris (changes requested):

glados-core/src/jsonrpc.rs
KolbyML: Ready for another look 🫡

mrferris (approved):
⭐️

KolbyML:
Thank you for the review @mrferris I really appreciate it

Issue #161, opened 2023-09-22
Add Radius and "Should store" column in Audit page


kdeme:
The audit page is very useful to track how content is available over the network and how many hops are done for a recursive find content.

Now when a `No Progress` node occurs, it would be nice to know whether that node should have had the content or not.

I think that to achieve this we could add a radius column that could optionally be filled in if the `Origin` node happens to know the radius of the `No Progress` (or other) nodes. Additionally, the calculation, based on distance and radius of this node, can be done to see whether this node should store this content in theory.  This could be indicated in another column, or perhaps be indicated with a color, e.g. marking the whole row red if it didn't have the content but should have had it.

Additional summaries could be made based on how many nodes failed to fulfill their task, etc.

I assume that the `portal_historyTraceRecursiveFindContent` call would need a slight adaptation to include the optional radius information.

edit: Actually, the information is equally useful for `Responded` nodes.

kdeme:
Since there  is the `is-content-in-deadzone` API since a while, a perhaps simpler option would be to have some special marking for any audit that failed and where the https://glados.ethdevops.io/api/is-content-in-deadzone/0x... returns no ENRs

PR #289, opened 2024-06-28
feat: add query trace node radius interpolation

mrferris:
Implements #161

**What was wrong?**

Query traces don't show any information about whether a given node should have had the data, according to its radius at the time of the trace.

**How was this fixed?**
A lookup is done to find in census data the radius for each node closest to the time at which the trace took place. This information is encoded within the `QueryTrace` object, and then visualized client-side.

Screenshot 2024-07-11 at 4 01 21 PM

### DIFF
--- a/glados-web/Cargo.toml
+++ b/glados-web/Cargo.toml
@@ -18,6 +18,7 @@ clap = { version = "4.0.26", features = ["derive"] }
 entity = { path = "../entity" }
 env_logger = "0.9.3"
 ethportal-api = { git = "https://github.com/ethereum/trin" }
+enr = "0.10.0"
 glados-core = { path = "../glados-core" }
 migration = { path = "../migration" }
 itertools = "0.10.5"
--- a/glados-web/assets/js/trace/main.js
+++ b/glados-web/assets/js/trace/main.js
@@ -1,7 +1,7 @@
 // Creates a graph from a trace object and returns an SVG.
 function createGraph(graphData) {
 
-    return ForceGraph(graphData, {
+    let graph = ForceGraph(graphData, {
         nodeId: d => d.id,
         nodeGroup: d => d.group,
         nodeGroups: [0, 1, 2, 3, 4, 5, 6, 7, 8, 9],
@@ -11,6 +11,8 @@ function createGraph(graphData) {
         height: $('#graph').height(),
         invalidation: null,
     });
+
+    return graph;
 
 }
 
@@ -52,6 +54,7 @@ function createGraphData(trace) {
       let distance = BigInt(meta.distance);
       let distanceLog2 = bigLog2(distance);
       let client = decodedEnr.client;
+      let radius = meta.radius;
 
       metadata[nodeId] = {
         enr,
@@ -59,7 +62,8 @@ function createGraphData(trace) {
         port,
         distance,
         distanceLog2,
-        client
+        client,
+        radius
       }
     });
 
@@ -197,6 +201,12 @@ function generateNodeMetadata(node) {
     if (client !== undefined) {
         metadata += `${client}`;
     }
+    const radius_numerator = BigInt(node.radius);
+    const radius_denominator = BigInt("0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff");
+    const radius_percentage = (Number(radius_numerator.toString()) / Number(radius_denominator.toString())) * 100;
+    if (node.radius){
+        metadata += `\nRadius Percent: ${radius_percentage.toFixed(2)}%`;
+    }
     return metadata;
 
 }
@@ -204,7 +214,6 @@ function generateNodeMetadata(node) {
 function generateTable(nodes) {
     nodes.sort((a, b) => a.distance < b.distance ? -1 : (a.distance > b.distance) ? 1 : 0);
     nodes.forEach((node, index) => {
-
         let nodeIdString = node.id;
         nodeIdString = nodeIdString.substr(0, 6)
             + '...' + nodeIdString.substr(nodeIdString.length - 4, nodeIdString.length);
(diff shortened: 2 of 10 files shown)


mrferris:
@kdeme Won't let me request review from you, but i'm looking for review from you on this. Let me know your thoughts.

kdeme (approved):
lgtm

glados-web/assets/js/trace/main.js
```diff
@@ -96,6 +104,8 @@ function createGraphData(trace) {
                 durationMs,
                 ...metadata[nodeId]
             });
+            console.log('asdfasdf');
+            console.log(metadata[nodeId]);
```
KolbyML, line 108:
```suggestion
```

These seem to be debug logs left in from development

glados-web/assets/js/trace/main.js
```diff
@@ -38,6 +42,8 @@ function createGraphData(trace) {
 
     let successfulRoute = computeSuccessfulRoute(trace);
 
+    console.log(trace);
+
```
KolbyML, line 46:
```suggestion
```

These seem to be debug logs left in from development

glados-web/Cargo.toml
```diff
@@ -18,6 +18,7 @@ clap = { version = "4.0.26", features = ["derive"] }
 entity = { path = "../entity" }
 env_logger = "0.9.3"
 ethportal-api = { git = "https://github.com/ethereum/trin" }
+discv5 = "0.4.1"
```
KolbyML, line 21: Since we are just using `sigp/enr` should we use it directly instead of using the re-exported instance of it. No `sigp/discv5` code is pulled in so I think just depending on `sigp/enr` itself, the only thing this PR uses should be the play

glados-web/src/routes.rs
```diff
@@ -838,12 +845,99 @@ pub async fn contentaudit_detail(
...
+                        FROM census_node cn2
+                        JOIN record r2 ON cn2.record_id = r2.id
+                        WHERE r2.node_id = n2.id
+                        ORDER BY cn2.surveyed_at - $2::timestamp
+                        LIMIT 1)
+                    FROM
+                        node n2
+                    WHERE
+                        n2.node_id = n.node_id
+                );
```
KolbyML, line 906: nit: using math notation (abbreviation to 1-2 letters) makes harder to read. Maybe instead of `n` -> `node_1` and `n2` -> `node_2` etc

I wouldn't say this is a blocker so up to you

glados-web/templates/contentaudit_detail.html
```diff
@@ -94,4 +111,4 @@
     });
 </script>
 
-{% endblock %}
+{% endblock %}
```
KolbyML, line 114: This file is missing an empty line at the end

glados-web/src/routes.rs
```diff
@@ -838,12 +845,99 @@ pub async fn contentaudit_detail(
...
+        };
+
+    // Add radius info to node metadata.
+    trace.metadata.iter_mut().for_each(|(node_id, node_info)| {
+        if let Some(radius) = nodes_with_radius.get(node_id) {
+            node_info.radius = Some(*radius);
+        }
+    });
+    // Update the trace with radius metadata.
+    audit.trace = serde_json::to_string(&trace).unwrap();
```
KolbyML, line 939: unwrap() used here

glados-web/src/routes.rs
```diff
@@ -838,12 +845,99 @@ pub async fn contentaudit_detail(
...
+    let trace_string = &audit.trace;
+    let mut trace: QueryTrace = serde_json::from_str(trace_string).unwrap();
+
+    // Get the timestamp of the query
+    let query_timestamp = trace.started_at_ms;
+    let timestamp: DateTime<Utc> = {
+        let duration = query_timestamp.duration_since(UNIX_EPOCH).unwrap();
+        Utc.timestamp_opt(duration.as_secs() as i64, duration.subsec_nanos() as u32)
+            .single()
+            .unwrap()
```
KolbyML, line 863: unwrap() used here

glados-web/src/routes.rs
```diff
@@ -838,12 +845,99 @@ pub async fn contentaudit_detail(
...
         .unwrap()
         .expect("No audit found");
 
+    let trace_string = &audit.trace;
+    let mut trace: QueryTrace = serde_json::from_str(trace_string).unwrap();
+
+    // Get the timestamp of the query
+    let query_timestamp = trace.started_at_ms;
+    let timestamp: DateTime<Utc> = {
+        let duration = query_timestamp.duration_since(UNIX_EPOCH).unwrap();
```
KolbyML, line 860: unwrap() used here

glados-web/src/routes.rs
```diff
@@ -838,12 +845,99 @@ pub async fn contentaudit_detail(
...
     info!("Audit ID: {}", audit_id);
-    let audit = content_audit::Entity::find_by_id(audit_id)
+    let mut audit = content_audit::Entity::find_by_id(audit_id)
         .one(&state.database_connection)
         .await
         .unwrap()
         .expect("No audit found");
 
+    let trace_string = &audit.trace;
+    let mut trace: QueryTrace = serde_json::from_str(trace_string).unwrap();
```
KolbyML, line 855: unwrap() used here

KolbyML (approved):
We should use `.expect(xyz)` instead of `.unwrap()`. An expect message makes it easier to find out what panicked when debugging. `unwrap()` is fine for tests, but we should avoid using in it in production code such as this.

I don't think we need to fix every instance of it in this PR, but we should fix it in the `unwrap()`'s added in this PR.

The long term goal should be to remove the use of `unwrap()` in production code and make the linter go ding ding ding when it does happen like we do in `ethereum/trin`

:shipit: Looks good I think this can be a useful feature and can possibly display issues with our bridging software which is very important to know about.

Issue #286, opened 2024-06-06
Graph Titles and Info tooltips on Dashboard


carver:
Add some basic graph necessities to everything on the Glados main dashboard:
- [x] Title
- [ ] Axis name
- [x] Tooltip giving more explanation of the purpose of the graph
- [x] ~Node's announced radius~ _Hm, the tooltip shows the radius in the chart, not sure what else this could mean. 🤷‍♂_ 
- [x] Colors that make sense (Network coverage colors should just be light/dark versions of the same color, and avoid confusion with existing client type colors)

mrferris:
Heads up the tooltip stuff was merged here: https://github.com/ethereum/glados/pull/282

PR #304, opened 2024-07-25
Dashboard chart titles

carver:
Checklist item on #286 -- I added some titles for the charts

image

Any thoughts on the names? I'm thinking it should be "Audit Success, by week" since you can page it.

---

Side note: I had to run an audit in order to generate data and make sure that the title didn't overlap. I didn't see a section in the readme to run the audit, so I figured it out and added a quick one. Any suggestions for how to improve that section?

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,6 +87,8 @@ $ cargo run -p glados-web -- --database-url DATABASE_URL
 
 This must be run from the project root, or static assets will fail to load, with 404 errors.
 
+You should then be able to view the web application at `http://127.0.0.1:3001/` in your browser.
+
 ### Running a census with `glados-cartographer`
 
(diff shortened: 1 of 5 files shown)


carver:
Added a few more tweaks to the coverage chart, see the latest commit message for details. Here's a pic:
image

mrferris (approved):
⛹️⭐️

Thoughts:

1.) On the scatterplot, I think something like "Node's Radius Percentage of Keyspace" might be more descriptive

2.) I agree, "Audit Success, by week" sounds good

3.) Instructions for running `glados-audit` are on the `SETUP_GUIDE.md`, good catch on putting them in the readme as well.

carver:
> 3.) Instructions for running `glados-audit` are on the `SETUP_GUIDE.md`, good catch on putting them in the readme as well.

Huh, that guide doesn't use the `--provider-url` flag, so I guess I'll take it out of the readme. Should we remove it as an option from the audit CLI?

Issue #301, opened 2024-07-02
Sort query trace nodes by node ID to visualize movement through network space


mrferris:
There should maybe be an option and the /audit/ page to turn this on or off.

carver:
Looks like this can be closed, because of #308 -- it's been fun to play with!

PR #308, opened 2024-08-13
feat: adds sorting of traces by node ID, with content ID marker

mrferris:
Implements #301 

Toggle-able via selector, with permanent memory of preferred selection.

Screenshot 2024-09-19 at 9 43 27 AM

### DIFF
--- a/glados-web/assets/js/trace/main.js
+++ b/glados-web/assets/js/trace/main.js
@@ -1,5 +1,5 @@
 // Creates a graph from a trace object and returns an SVG.
-function createGraph(graphData) {
+function createGraph(graphData, sortByNodeId = false) {
 
     let graph = ForceGraph(graphData, {
         nodeId: d => d.id,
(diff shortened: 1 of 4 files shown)


KolbyML:
Does the "should store" outline not work on this graph? From the picture in the PR it doesn't show the "should store" outline anywhere which is why I am asking

glados-web/assets/js/trace/main.js
```diff
@@ -146,8 +147,10 @@ function createGraphData(trace) {
         metadata: {
             nodesContacted: nodes.length - 1,
             nodesResponded: Object.keys(responses).length - 1
-        }
+        },
+        contentId: '0x' + trace.targetId.map(byte => byte.toString(16).padStart(2, '0')).join(''),
     }
+    console.log(graph);
```
KolbyML, line 153: Do we still need this debug log?

KolbyML (commented):
Most things look good, I am interested in a comment if "should store" is broken or not, but other then that the PR looks very good

glados-web/assets/js/trace/main.js
mrferris: I removed it previously and have found that it's actually nice to have so I've added it back.

mrferris:
@KolbyML It's missing from the screenshot because this is from a local instance in which I hadn't run any censuses, which is where the "should store" data comes from. "should store" still works, and it happens on the backend which isn't touched by this PR.

KolbyML (approved):
:shipit:

Issue #293, opened 2024-07-02
Census feature parity across history, state, and beacon networks


mrferris:
Glados's cartographer currently only runs a census on the history network. 

It should also run/log censuses for the state and beacon networks. 

This should be likely be done with a `--network` flag on the `glados-cartographer` cli.

PR #309, opened 2024-08-15
feat: Adds cartographer support for beacon and state networks

mrferris:
Adds `--subnetwork history|beacon|state` flag to cartographer

Subsequent PR will add census info to `glados-web`

#293


KolbyML (approved):
:shipit: looks good

PR #320, opened 2024-10-01
feat: adds state audits & audit dashboard state filtering

mrferris:
This PR:

- activates population of state root data for blocks 0 - 1 million via a new `glados_monitor_state` container in `docker-compose.yml`
- activates auditing of state root data via a new `glados_audit_state` container in `docker-compose.yml`
- implements toggling of the audit dashboard between history and state

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,8 +37,34 @@ services:
       - "host.docker.internal:host-gateway"
     restart: always
 
+  glados_audit_state:
+    command: "--database-url postgres://postgres:${GLADOS_POSTGRES_PASSWORD}@glados_postgres:5432/glados --portal-client http://host.docker.internal:8545 --state --history=false"
+    image: portalnetwork/glados-audit:latest
+    environment:
+      RUST_LOG: warn,glados_audit=info
+    depends_on:
+      - glados_postgres
+      - portal_client
+    networks:
+      - glados-net
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+    restart: always
+
   glados_monitor:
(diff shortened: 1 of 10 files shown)


pipermerriam (commented):
:rocket:

docker-compose.yml
```diff
@@ -49,6 +63,18 @@ services:
...
+    image: portalnetwork/glados-monitor:latest
+    environment:
+      RUST_LOG: warn,glados_monitor=info
+    depends_on:
+      - glados_audit
+      - glados_postgres
+    networks:
+      - glados-net
+    restart: always
+
```
KolbyML, line 77: We don't support latest state yet, this should be set to import the state_roots from 0 to 1 million

```suggestion
  glados_monitor_state:
    command: "--database-url postgres://postgres:${GLADOS_POSTGRES_PASSWORD}@glados_postgres:5432/glados populate-state-roots-range --start-block-number 0 --end-block-number 1000001 --provider-url ${GLADOS_PROVIDER_URL?Glados monitor provider URL required}"
    image: portalnetwork/glados-monitor:latest
    environment:
      RUST_LOG: warn,glados_monitor=info
    depends_on:
      - glados_audit
      - glados_postgres
    networks:
      - glados-net
    restart: always
```

Without this change the audits will just fail. We will enable `follow-head-state` after we have bridges gossiping latest

KolbyML (commented):
We can't enable `follow-head-state` yet as we don't gossip latest state data onto the network yet. This will cause all audits to fail. Instead we need to use `populate-state-roots-range --start-block-number 0 --end-block-number 1000001` which will make it so state audit only audits state we are actively gossiping on the network

KolbyML (approved):
:shipit:

PR #321, opened 2024-10-02
Build correctly, with the latest ethportal-api

carver:
(I think) I need the latest ethportal-api to parse the new query trace. I tried to force cargo to use the latest, but then glados-web wouldn't build for me locally.

~To start, this PR is making sure I don't just have a local build problem. After confirming that, I'll switch over to making glados build again.~

### DIFF
--- a/glados-audit/src/state.rs
+++ b/glados-audit/src/state.rs
@@ -193,9 +193,12 @@ async fn random_state_walk(
 
         let encoded_trie_node: EncodedTrieNode = content_value.to_vec().into();
 
-        let trie_node = encoded_trie_node
-            .as_trie_node()
-            .expect("Trie node received from the portal network should be decoded as a trie node");
+        let trie_node = encoded_trie_node.as_trie_node().map_err(|err| {
+            (
+                anyhow!("Error decoding node while walking trie: {err:?}"),
+                current_content_key.clone(),
+            )
+        })?;
         match process_trie_node(current_content_key, trie_node).await? {
             (next_content_key, false) => {
                 current_content_key = next_content_key;
--- a/glados-audit/src/state.rs
+++ b/glados-audit/src/state.rs
@@ -13,9 +13,10 @@ use eth_trie::node::Node;
 use ethportal_api::{
     jsonrpsee::http_client::HttpClient,
     types::{
-        content_key::state::AccountTrieNodeKey, state::ContentInfo, state_trie::nibbles::Nibbles,
+        content_key::state::AccountTrieNodeKey, portal::ContentInfo, state_trie::nibbles::Nibbles,
+        state_trie::EncodedTrieNode,
     },
-    StateContentKey, StateContentValue, StateNetworkApiClient,
+    StateContentKey, StateNetworkApiClient,
 };
 use glados_core::{db::store_content_key, jsonrpc::PortalClient};
 use rand::seq::IteratorRandom;
@@ -190,26 +191,9 @@ async fn random_state_walk(
             }
         };
 
-        let encoded_trie_node = match content_value {
-            StateContentValue::TrieNode(encoded_trie_node) => encoded_trie_node,
-            other_state_content_value => {
-                return Err((anyhow!(
-                        "State random walk audit recevied unexpected content type: {other_state_content_value:?}"
-                    ), current_content_key));
-            }
-        };
-
-        if encoded_trie_node.node.node_hash() != current_content_key.node_hash {
-            return Err((
-                anyhow!(
-                    "State random walk audit recevied unexpected node hash: {current_content_key:?} {encoded_trie_node:?}"
-                ),
-                current_content_key,
-            ));
-        }
+        let encoded_trie_node: EncodedTrieNode = content_value.to_vec().into();
 
         let trie_node = encoded_trie_node
-            .node
             .as_trie_node()
             .expect("Trie node received from the portal network should be decoded as a trie node");
         match process_trie_node(current_content_key, trie_node).await? {
--- a/glados-audit/src/validation.rs
+++ b/glados-audit/src/validation.rs
@@ -115,9 +115,5 @@ fn validate_history(content_key: &HistoryContentKey, content_bytes: &[u8]) -> bo
             warn!("Need to call trusted provider to check receipts correctness.");
             true
         }
-        HistoryContentValue::EpochAccumulator(_e) => {
-            warn!("Need to check epoch master accumulator for correctness.");
-            true
-        }
     }
 }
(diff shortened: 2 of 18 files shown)


glados-audit/src/state.rs
```diff
@@ -190,26 +191,9 @@ async fn random_state_walk(
...
-
-        if encoded_trie_node.node.node_hash() != current_content_key.node_hash {
-            return Err((
-                anyhow!(
-                    "State random walk audit recevied unexpected node hash: {current_content_key:?} {encoded_trie_node:?}"
-                ),
-                current_content_key,
-            ));
-        }
+        let encoded_trie_node: EncodedTrieNode = content_value.to_vec().into();
```
morph-dev, line 194: nit: I think this should work, but I don't think this is the best way of doing things (because it can panic if something is wrong). Instead, I would:

1. Have `let content_key = StateContentKey::AccountTrieNode(current_content_key.clone());` at the beginning of loop (line 164) and reuse it later (line 166).

2. replace this and following line (196) with:
```rust
let trie_node = match StateContentValue::decode(&content_key, &content_value)
    .map_err(|err| anyhow!("Error decoding content value: {err}"))
    .and_then(|content_value| match content_value {
        StateContentValue::TrieNode(trie_node) => Ok(trie_node.node),
        _ => Err(anyhow!("Wrong content value type, expected: TrieNode")),
    })
    .and_then(|encoded_trie_node| Ok(encoded_trie_node.as_trie_node()?))
{
    Ok(trie_node) => trie_node,
    Err(err) => return Err((err, current_content_key)),
};
```

glados-audit/src/validation.rs
```diff
@@ -1,15 +1,16 @@
...
 use ethportal_api::utils::bytes::hex_encode;
-use ethportal_api::{BeaconContentKey, BlockHeaderKey, HistoryContentKey, OverlayContentKey};
+use ethportal_api::{BeaconContentKey, HistoryContentKey, OverlayContentKey, RawContentKey};
 use ethportal_api::{BeaconContentValue, ContentValue, HistoryContentValue};
 use tracing::warn;
 
 /// Checks that content bytes correspond to a correctly formatted
 /// content value.
 pub fn content_is_valid(content: &content::Model, content_bytes: &[u8]) -> bool {
+    let raw_key = RawContentKey::from(content.content_key.clone());
```
morph-dev, line 10: nit: no big deal, but you can also do: `let raw_key = RawContentKey::from_iter(&content.content_key);`

glados-core/src/jsonrpc.rs
```diff
@@ -172,15 +168,16 @@ impl PortalApi {
         self,
         content: &content::Model,
     ) -> Result<Option<Content>, JsonRpcError> {
+        let raw_key = RawContentKey::from(content.content_key.clone());
```
morph-dev, line 171: nit: also here

glados-web/src/routes.rs
```diff
@@ -486,6 +486,7 @@ pub async fn contentkey_detail(
             StatusCode::NOT_FOUND
         })?;
 
+    let content_key_raw: RawContentKey = content_key_raw.into();
```
morph-dev, line 489: nit: you don't have to convert it to `RawContentKey`.

Instead, you can use `HistoryContentKey::from_hex(&content_key_hex)` below (instead of `HistoryContentKey::try_from(content_key_raw.clone())`).

Same for state and beacon.

morph-dev (approved):
In general, looks good to me.

I think all my comments (except one) are small suggestion on how type conversion could have been done differently. Not necessarily because I think it's a better way, but because I wanted to provide alternative.

The only bigger comment is related to state audit and not strictly related to this PR. Therefore, while nice to have, it can be ignored and added in future improvements.

glados-audit/src/validation.rs
carver: Yup, who doesn't love dropping an unnecessary clone?

glados-web/src/routes.rs
carver: Yeah, the only downside of this is that we want to understand the difference between a key that isn't valid hex-bytes, and a key that doesn't match any valid content key pattern. I think I see a reasonable way to do that. It's a little slower in the "invalid hex" path, but that seems okay.

glados-audit/src/state.rs
carver: Ok, I agree that we should get rid of the panic. Even though the `.expect("Trie node received from the portal network should be decoded as a trie node")` was not introduced by this PR, I'm happy to fix that broken window along the way.

The shortest version I settled on was:

A)
```rs
let encoded_trie_node: EncodedTrieNode = content_value.to_vec().into();        
let trie_node = encoded_trie_node.as_trie_node().map_err(|err| {        
    (        
        anyhow!("Error decoding node while walking trie: {err:?}"),        
        current_content_key.clone(),        
    )        
})?; 
```

Versus the touched-up version from your comment:

B)
```rs
let full_key = StateContentKey::AccountTrieNode(current_content_key.clone());        
let trie_node = match StateContentValue::decode(&full_key, &content_value)        
    .map_err(|err| anyhow!("Error decoding content value: {err}"))        
    .and_then(|content_value| match content_value {        
        StateContentValue::TrieNode(trie_node) => Ok(trie_node.node),        
        other => Err(anyhow!("Wrong content value type, expected: TrieNode, got: {other:?}")),        
    })        
    .and_then(|encoded_trie_node| Ok(encoded_trie_node.as_trie_node()?))        
{        
    Ok(trie_node) => trie_node,        
    Err(err) => return Err((err, current_content_key)),        
}; 
```

The most notable tradeoffs I see are:
A
- Shorter code
- It assumes the type of the state content value is an account trie node

B
- Longer code
- More generalized, maybe laying the groundwork for auditing things like the bytecode

I can imagine a rationale for B could be laying some groundwork toward supporting bytecode and contract storage content values. Though I think in this PR which is focused on getting glados working with the latest ethportal-api, it feels a bit tangential. It would probably make sense to make broader changes in `process_trie_nodes()`, etc, if we're really going to take a stab at generalizing.

In the context it's in, A seems like it's leaving the code better than I found it, without getting too invested in generalizing to handling other content values. So I guess I'm leaning toward A right now. Do you see anything I'm missing?

glados-web/src/routes.rs
carver: I think there are more optimized approaches, but I didn't like how any of my attempts went. So I just went with the most naive change that seems fine.

It does some things that make me cringe a tiny bit, like decoding the hex value for the DB, then using the original hex value, which is getting decoded over and over again for each key check. But it's not a big deal, and I think it's time to just call it a distraction and move on.

PR #329, opened 2024-10-18
chore: new ethportal-api with recursive_find -> get

carver:
The json-rpc api changed calls with recursive_find_content to get_content (and the same with trace).

Update to get the latest ethportal-api with the changes.


KolbyML (approved):
:shipit:

PR #330, opened 2024-10-19
chore: update ethportal-api which updates ContentInfo return types

KolbyML:
updates the version of ethportal-api this update removes code paths which weren't included in the JSON-RPC return types like content-id which is an internal trin thing.

### DIFF
--- a/glados-audit/src/state.rs
+++ b/glados-audit/src/state.rs
@@ -177,12 +177,7 @@ async fn random_state_walk(
             }
         };
 
-        let GetContentInfo {
-            content: content_value,
-            ..
-        } = response;
-
-        let encoded_trie_node: EncodedTrieNode = content_value.to_vec().into();
+        let encoded_trie_node: EncodedTrieNode = response.content.to_vec().into();
 
         let trie_node = encoded_trie_node.as_trie_node().map_err(|err| {
             (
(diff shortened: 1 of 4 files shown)


glados-core/src/jsonrpc.rs
```diff
@@ -173,10 +173,9 @@ impl PortalApi {
             content::SubProtocol::History => {
                 match HistoryNetworkApiClient::get_content(&self.client, raw_key.try_into()?).await
                 {
-                    Ok(ContentInfo::Content { content, .. }) => Ok(Some(Content {
+                    Ok(GetContentInfo { content, .. }) => Ok(Some(Content {
```
morph-dev, line 176: nit: just style preference, so completely up to you. But this can be done as:
```rust
Ok(content_info) => Ok(Some(Content {
    raw: content_info.content.into();
}))
```

As I said completely up to you. But if you change it, change it below as well.

morph-dev (approved):
:shipit:

glados-audit/src/state.rs
```diff
@@ -176,18 +177,10 @@ async fn random_state_walk(
...
-                ..
-            } => content_value,
-            other_content_info => {
-                return Err((
-                    anyhow!("Error unexpected get_content response: {other_content_info:?}"),
-                    current_content_key,
-                ));
-            }
-        };
+        let GetContentInfo {
```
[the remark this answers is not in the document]
KolbyML: good call I forgot you could do this

glados-core/src/jsonrpc.rs
KolbyML: this is cleaner yeah

Issue #297, opened 2024-07-02
API endpoint for getting failed content keys


mrferris:
Glados should have an API endpoint and a simple web dashboard feature for getting a list of content keys that had failed recently. This is useful for both manual debugging and for feeding into a bridge node to intelligently plug holes in the dataset. 

Previously I made a script for nick to do this for manually debugging:

`ssh glados "psql postgres://postgres:password@localhost:5432/glados -c \"\\copy (SELECT DISTINCT c.content_key FROM content c JOIN content_audit ca ON c.id = ca.content_key WHERE ca.result = 0 AND ca.strategy_used = 5 AND ca.created_at >= CURRENT_TIMESTAMP - INTERVAL '24 hours') TO '/home/devops/failed_4444s.txt';\"" && scp glados:/home/devops/failed_4444s.txt ~/failed_4444s.txt`

Ktl-XV:
@mrferris can I take this issue?

PR #333, opened 2024-10-24
feat: Adds `/api/failed-keys/` endpoint

Ktl-XV:
Adds an endpoint exposing the content keys of audits that have failed in the last 24 hours. https://github.com/ethereum/glados/issues/297

### DIFF
--- a/entity/src/content_audit.rs
+++ b/entity/src/content_audit.rs
@@ -2,12 +2,13 @@
 use crate::content;
 use crate::utils;
 use anyhow::{bail, Result};
-use chrono::{DateTime, Utc};
+use chrono::{DateTime, TimeDelta, Utc};
 use clap::ValueEnum;
-use ethportal_api::OverlayContentKey;
+use content::SubProtocol;
+use ethportal_api::{utils::bytes::hex_encode, OverlayContentKey};
 use sea_orm::{
-    entity::prelude::*, strum::IntoEnumIterator, ActiveValue::NotSet, DeriveActiveEnum, Set,
-    TryGetable,
+    entity::prelude::*, strum::IntoEnumIterator, ActiveValue::NotSet, Condition, DeriveActiveEnum,
+    JoinType, QueryOrder, QuerySelect, Set, TryGetable,
 };
 use sea_query::{ArrayType, Nullable, SeaRc, ValueType, ValueTypeErr};
 
@@ -58,6 +59,21 @@ impl From<i32> for HistorySelectionStrategy {
     }
 }
 
+impl TryFrom<String> for HistorySelectionStrategy {
+    type Error = anyhow::Error;
+    fn try_from(value: String) -> Result<Self, Self::Error> {
+        match value.as_str() {
+            "Latest" => Ok(HistorySelectionStrategy::Latest),
+            "Random" => Ok(HistorySelectionStrategy::Random),
+            "Failed" => Ok(HistorySelectionStrategy::Failed),
+            "SelectOldestUnaudited" => Ok(HistorySelectionStrategy::SelectOldestUnaudited),
+            "SpecificContentKey" => Ok(HistorySelectionStrategy::SpecificContentKey),
+            "FourFours" => Ok(HistorySelectionStrategy::FourFours),
+            _ => bail!("Invalid value for HistorySelectionStrategy {}", value),
+        }
+    }
+}
+
 #[derive(Debug, Clone, Eq, Hash, PartialEq, EnumIter, DeriveActiveEnum, ValueEnum)]
 #[clap(rename_all = "snake_case")]
 #[sea_orm(rs_type = "i32", db_type = "Integer")]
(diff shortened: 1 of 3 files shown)


glados-web/src/routes.rs
```diff
@@ -809,6 +809,38 @@ pub async fn get_audit_stats_handler(
...
+        None => 1,
+        Some(page) => page.parse::<u32>().unwrap_or(1),
+    };
+
+    let failed_keys =
+        content_audit::get_failed_keys(subprotocol, strategy, page, &state.database_connection)
+            .await
+            .map_err(|e| {
+                error!(err=?e, "Could not fetch failed keys");
+                StatusCode::INTERNAL_SERVER_ERROR
```
Ktl-XV, line 838: When the user sends an invalid network/strategy combination, the server returns status 500 instead of 400, and the offending field is logged, not included in the response. Should these errors be included in the response?

entity/src/content_audit.rs
```diff
@@ -335,6 +371,46 @@ pub async fn get_audits<T: OverlayContentKey>(
...
+    let strat: SelectionStrategy = match subprotocol {
+        SubProtocol::History => SelectionStrategy::History(strategy_used.try_into()?),
+        SubProtocol::State => SelectionStrategy::State(strategy_used.try_into()?),
+        SubProtocol::Beacon => SelectionStrategy::Beacon(strategy_used.try_into()?),
+    };
+
+    let twenty_four_hours = TimeDelta::try_hours(24).expect("Couldn't calculate 24 hours delta.");
+    let twenty_four_hours_ago = Utc::now() - twenty_four_hours;
+
+    let keys = Entity::find()
```
Ktl-XV, line 391: Pagination and uniqueness do not mix well with dynamic time ranges, since items might be pushed to different pages in between requests.

Since the client will need to duplicate on their side, I have removed the distinct to make the query simpler.

KolbyML:
@Ktl-XV Is this PR still draft or is it ready for review?

Ktl-XV:
I'm waiting on feedback from @mrferris on the questions I included in the PRs comments, but it is ready. I'll mark it as such, could you take a look?

glados-web/src/routes.rs
mrferris: I think it's fine to leave it in the logs

entity/src/content_audit.rs
mrferris: Can you give an example of uniqueness causing a problem across pages? The intention of the `DISTINCT` is just to remove duplicate content keys because there can be multiple failed audits for the same content key.

entity/src/content_audit.rs
```diff
@@ -335,6 +371,46 @@ pub async fn get_audits<T: OverlayContentKey>(
...
+        .filter(
+            Condition::all()
+                .add(Column::Result.eq(AuditResult::Failure))
+                .add(Column::StrategyUsed.eq(strat))
+                .add(Column::CreatedAt.gt(twenty_four_hours_ago)),
+        )
+        .order_by_asc(Column::Id)
+        .limit(page_size as u64)
+        .offset(((page - 1) * page_size) as u64)
+        .into_tuple()
```
mrferris, line 404: Let's replace this SeaQL expression with raw SQL using SeaQL's `from_sql_and_values` (search that function within glados repo to see usage examples). We want to get away from using SeaQL like this because it's so much less standard and less documented than SQL itself.

entity/src/content_audit.rs
Ktl-XV: After thinking more about it, the core problem is not actually the use of `DISTINCT` but trying to paginate on a moving window, where a new item might come in (or out of the 24h window if sorted `ASC`) in between page requests, shifting items to different pages.

1. One option can be to let the client select the time range they want to query (increasing complexity for the client, and adding a delay since the start of the range has to be enough time in the past [likely a couple of seconds] to account for out of sync clocks)
2. Deal with the skipped keys IDs by having a large enough page size so pagination is usually not needed
3. If the client runs frequently enough, and not many keys fail, any skipped key will be caught in the subsequent runs

I prefer option 1 since it is the most robust, and a couple of seconds of delay is probably good enough, but is it OK to ask the client to supply a time range?

entity/src/content_audit.rs
Ktl-XV: Replaced SeaQL with raw SQL

entity/src/content_audit.rs
```diff
@@ -335,6 +376,56 @@ pub async fn get_audits<T: OverlayContentKey>(
         .await?)
 }
 
+pub async fn get_failed_keys(
+    subprotocol: SubProtocol,
+    strategy_used: String,
+    page: u32,
+    conn: &DatabaseConnection,
+) -> Result<Vec<String>> {
+    let page_size: u32 = 1000;
```
mrferris, line 385: small semantic thing, but it's good to define constants as constants:
```suggestion
    const PAGE_SIZE: u32 = 1000;
```

entity/src/content_audit.rs
```diff
@@ -335,6 +376,56 @@ pub async fn get_audits<T: OverlayContentKey>(
...
+    let subprotocol_strategy: SelectionStrategy = match subprotocol {
+        SubProtocol::History => SelectionStrategy::History(strategy_used.try_into()?),
+        SubProtocol::State => SelectionStrategy::State(strategy_used.try_into()?),
+        SubProtocol::Beacon => SelectionStrategy::Beacon(strategy_used.try_into()?),
+    };
+
+    let keys_result = FailedKeysResult::find_by_statement(Statement::from_sql_and_values(
+        DbBackend::Postgres,
+        "
+        SELECT
```
mrferris, line 396: This query is equivalent (I verified via `explain analyze` that it generates the same plan and runs the same) and I think a little more readable without the sub-query:
```suggestion
        SELECT
              content.content_key
        FROM content_audit
        INNER JOIN content ON content.id = content_audit.content_key
        WHERE 
            content_audit.result = 0 AND
            content_audit.strategy_used = $1 AND
            content_audit.created_at >= CURRENT_TIMESTAMP - INTERVAL '24 hours'
        GROUP BY content.content_key
        ORDER BY MAX(content_audit.created_at) DESC
        LIMIT $2
        OFFSET $3
```

mrferris (approved):
Ready to merge once those changes are made

entity/src/content_audit.rs
mrferris: I think that having the moving window is ok for now, we can adjust based on user feedback

PR #335, opened 2024-10-31
fix: decode account trie node values correctly

mrferris:
State audits are failing due to 

`[2024-10-31T15:58:27Z ERROR glados_audit::state] Error during state audit. err=Error decoding node while walking trie: InvalidData content_key=AccountTrieNodeKey { path: Nibbles { nibbles: [] }, node_hash: 0x7f485a9dd816bcbfa98f6fa2334af80fabc71ac14a605685bde325c9eb1b5f9e }`

This was happening due to the usage of `ethportal_api::EncodedTrieNode` instead of `StateContentValue::decode` as is done within trin: https://github.com/ethereum/trin/blob/238e0b0b448f39409c34c8017c10f8a4212daee7/rpc/src/evm_state.rs#L169

### DIFF
--- a/glados-audit/src/state.rs
+++ b/glados-audit/src/state.rs
@@ -12,11 +12,8 @@ use entity::{
 use eth_trie::node::Node;
 use ethportal_api::{
     jsonrpsee::http_client::HttpClient,
-    types::{
-        content_key::state::AccountTrieNodeKey,
-        state_trie::{nibbles::Nibbles, EncodedTrieNode},
-    },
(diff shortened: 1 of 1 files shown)


carver (approved):
Yeah, this begs for a new ethportal-api approach. Something like:
```PortalRpcClient::get_content_value(key: StateContentKey) -> StateContentValue```

This seems like a very common usage pattern. Better than asking users to always `get_content()` and then `decode()`.

Before we change ethportal-api, this looks like the best solution 👍🏻

PR #340, opened 2024-12-09
chore: upgrade ethportal-api to latest

carver:
It uses a more recent version of eth-trie, which causes conflict errors with the locally imported version. Updating the local version is apparently trivial.

This is necessary to use the latest versions of trin.


njgheorghita (approved):
:ship:

PR #342, opened 2024-12-10
chore: remove provider from beacon monitor docker

carver:
No idea why this is done. I found it in a patch on the production server.

It seems like it should work. Mike, can you share context?


mrferris (approved):
Good catch, the hot-fix had simply been because `--provider-url` is not a valid flag.

PR #346, opened 2024-12-12
Update to ethportal v0.4.0

carver:


carver:
I suppose like we could skip this entirely. It doesn't trigger any changes, I guess because we aren't any related beacon endpoints.

But also I guess it doesn't hurt. 🤷🏻‍♂️

mrferris (approved):

Issue #336, opened 2024-11-16
Compute aggregate transfer success rate stats from the trace


carver:
Derive some statistics, by analyzing the query traces from all different audit strategies.

Derive the transfer fail rate as a percentage of attempted transfers. Break that down by:
- client type & version
- failure type (utp vs validation, etc)
- content type

We also probably want to keep a list of recent failures, per client. It doesn't need to be exhaustive. Just a few examples per client is sufficient (maybe per client version).

### Outline of remaining work
- [x] migration of sql
	- [ ] add new record field to internal transfer failure
	- [ ] convert existing failures from node id to record id
	- [ ] remove old sender_node field from internal transfer failure
- [x] migrate backwards
	- [ ] add sender_node back to internal failure
	- [ ] generate node id from record id
	- [ ] drop record id field
- [x] for local testing:
	- [ ] generate some transfer failures, using the old (node- [ ]based) method
	- [ ] run a census? (can we skip this step, or do we need it for records and key values?)
	- [ ] run a migration
	- [ ] check the record id matches the previous node id
	- [ ] run a backwards migration
	- [ ] check the node id matches itself from before
- [x] while recording new internal failures, find the record to link to, based on the most recent sequence number
- [ ] create a local query that generates the client-groupings of failures by day, by strategy
- [ ] deploy changes
- [ ] run query manually on production to see the stats by client
- [ ] look for indices to add to optimize query, to prepare to add it in the dashboard
- [ ] build a chart out of the stats grouped by client

carver:
First chart:

![Image](https://github.com/user-attachments/assets/ab279287-4ca8-4c65-b865-8a521d60b72c)

Then repeat that, but for each type of failure, like validation.

Really looking to channel the home page chart for audit success. Since I think it would be good to break down by client from the start, which means multiple data lines, we probably don't want multiple lines for each strategy. So maybe more like a radio button than a checkbox list. But that can be addressed down the line.

carver:
Actually, might start with just an absolute number of failures to start. That's quicker/easier to build, I expect.

carver:
Next up is a refactor & sql migration of transfer failures, after the learnings from manually aggregating live statistics.

PR #355, opened 2025-01-16
refactor(core): give direct access to QueryTrace

carver:
This will be useful for inspecting the results of the trace, and selectively recording them to the database. For example, if some transfers fail during the trace (even if one ultimately succeeds), it's useful to track those intermediate failures.

(Some precursor work for #336 )

### DIFF
--- a/glados-core/src/jsonrpc.rs
+++ b/glados-core/src/jsonrpc.rs
@@ -4,6 +4,7 @@ use alloy_primitives::hex::FromHexError;
 use entity::content;
 use ethportal_api::types::enr::Enr;
 use ethportal_api::types::portal::TraceContentInfo;
+use ethportal_api::types::query_trace::QueryTrace;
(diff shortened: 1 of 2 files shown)


glados-core/src/jsonrpc.rs
```diff
@@ -97,16 +97,32 @@ pub enum JsonRpcError {
...
+                if err_data.is_none() {
+                    return JsonRpcError::ContentNotFound { trace: None };
+                } else {
+                    let trace_str = err_data.unwrap().to_string();
+                    let trace = serde_json::from_str(&trace_str);
+                    if let Ok(trace) = trace {
+                        return JsonRpcError::ContentNotFound { trace: Some(trace) };
+                    } else {
+                        return JsonRpcError::InvalidQueryTrace(trace_str);
+                    }
```
carver, line 124: This feels unnecessarily verbose. I'll take another look before requesting a review. But I think the rest is straightforward.

PR #362, opened 2025-01-27
Add Database entries for every transfer failure

carver:
~~TODO: rebase on #355~~

Next major step toward #336 -- saving each transfer failure into the database.


glados-audit/src/lib.rs
```diff
@@ -448,6 +458,30 @@ async fn perform_single_audit(
...
+async fn create_entry_for_failures(
+    audit: content_audit::Model,
+    trace: QueryTrace,
+    conn: &DatabaseConnection,
+) {
+    // Create a list of the failures from the parsed trace json
+    for (sender_node_id, failure_entry) in trace.failures.into_iter() {
+        let fail_type = failure_entry.failure;
+        // For now, just log out the failure, and related info.
+        // Later, write this data into a new table in the database.
```
mrferris, line 471: Is this a leftover? Seems like the failure is being written into the database below.

glados-audit/src/lib.rs
```diff
@@ -475,3 +509,147 @@ async fn display_history_audit_result(
...
+            node: Set(Some(node.id)),
+        };
+        content_audit_active_model.insert(&conn).await?;
+
+        let loaded_audit = content_audit::Entity::find()
+            .filter(content_audit::Column::ContentKey.eq(1))
+            .one(&conn)
+            .await?
+            .unwrap();
+        assert_eq!(loaded_audit.result, AuditResult::Success);
```
mrferris, line 651: nit: This and the following line seem a little arbitrary, why not also test `strategy_used` and `content_key`? No need to address, just thinking out loud.

migration/src/m20250130_042751_create_audit_internal_failures.rs
```diff
@@ -0,0 +1,87 @@
...
+                            .not_null(),
+                    )
+                    .foreign_key(
+                        ForeignKey::create()
+                            .name("FK_auditinternalfailure_sender_node")
+                            .from(
+                                AuditInternalFailure::Table,
+                                AuditInternalFailure::SenderNode,
+                            )
+                            .to(Node::Table, Node::Id),
```
mrferris, line 46: I notice you don't define `on_delete` & `on_update` here, is that what you intended?

mrferris (approved):
Cool, can't wait to be able to see this info. 

One thought: it will be nice in the case of `InvalidContent` failures to be able to actually see the invalid content. Out of scope for this PR, just wondering if you have thoughts. That invalid data might be a bit big to just stuff in another column, but maybe i'm overthinking.

glados-audit/src/lib.rs
carver: I guess because it's the focus of the test, to get a trace out of a successful audit, that shows failed transfers. So it's validating those features of the audit are in place.

migration/src/m20250130_042751_create_audit_internal_failures.rs
carver: Yeah, it was certainly intended for on_delete. I was sort of considering them a package deal, but I suppose the on_update cascade makes sense to add. Here's my logic:

If we delete a Node record, I would rather keep the transfer failure info in the DB, even though it is missing the relevant node details. (Compare the situation to what happens if we delete an audit record: it's mostly useless to keep the transfer failure around without knowing what was audited, so a delete cascade makes sense)

In the extremely unlikely situation that we change the primary key of the node record, I guess we would want the failed transfer record to update the foreign key. So it's more correct to add the update cascade, even though I think it will never come up in practice.

PR #364, opened 2025-01-30
chore: upgrade sea-orm to 1.1.0

carver:


carver:
Hm, I was hoping this would be more straightforward. I've spent a bit of time on it, and I'm not sure how much more time it would take. The errors are pretty opaque to me.

PR #367, opened 2025-02-06
fix: use wrong seq when find Enr

### LINKED ISSUE #366: Shisui doesn't show up on glados with timestamp seq number enabled
Image

fearlessfe:
fix #366 
find Enr and insert Enr use different seq, so it will cause the insert error, only the seq bigger than i32::MAX will occor the error.


KolbyML (approved):
:shipit: looks good

PR #369, opened 2025-02-18
docs: fix for temp db in local test run

carver:
Database tests couldn't run locally, because they couldn't find initdb. This change adds a section to the readme, about how to diagnose and fix this particular issue.


mrferris (approved):
⭐️

PR #373, opened 2025-02-26
perf: speed up audit select queries

carver:
Use the existing indices more effectively by:
- using a left join instead of inner join
- move the protocol_id condition into the join on the content table

Both of these help avoid full table scans, which can shave many seconds off of the longest queries.


carver:
I tested them manually by running `explain analyze` on the hosted service to see how the total query time changed (and see where indices were correctly used).

mrferris (approved):

PR #386, opened 2025-04-02
fix(web): allow for census with no online nodes

Ktl-XV:
Currently a census where all nodes were offline (or Glados' Portal node couldn't find any) breaks the census explorer request (and view) for that 24 hour period, this PR will instead display it as a census with no online nodes.

### DIFF
--- a/glados-web/src/routes.rs
+++ b/glados-web/src/routes.rs
@@ -949,10 +949,10 @@ pub async fn census_explorer_list(
 
 #[derive(Debug, Clone, FromQueryResult)]
 pub struct NodeStatus {
-    enr_id: i32,
+    enr_id: Option<i32>,
     census_time: DateTime<Utc>,
     census_id: i32,
-    node_id: Vec<u8>,
+    node_id: Option<Vec<u8>>,
     present: bool,
 }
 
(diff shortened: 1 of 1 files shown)


carver (commented):
It's my first time in this code, but I don't see anything too crazy.

Silly question: where are `enr_id` and `node_id` being set? Is there some magic in the `FromQueryResult` that knows to populate the struct differently when there's an `Option`? How was it populating before if the ENR was missing?

Ktl-XV:
> Silly question: where are enr_id and node_id being set?

They are being set in `NodeStatus::find_by_statement`. 


> Is there some magic in the FromQueryResult that knows to populate the struct differently when there's an Option?
> How was it populating before if the ENR was missing?

Yes, the Option allows for nullable fields. Currently without it, when trying to decode a `NULL` into `enr_id` the request errors out with `A null value was encountered while decoding "enr_id"` returning a 500

carver:
> > Silly question: where are enr_id and node_id being set?
> 
> They are being set in `NodeStatus::find_by_statement`.

Ah, that's where they are configured in requesting the data. I meant: where are they being assigned into the database? But it's ok, I think I understand now that there was just a mismatch between the struct definition and the actual sql definition. I just thought that wasn't permitted to compile by seaql-orm, but now I think it is allowed.

PR #387, opened 2025-04-04
feat: in transfer failure records, link to ENR instead of node

carver:
Working toward #336 

Previously, audit_internal_failure tracked the node_id of the sending node in each failure row. This left the precise information about the sender ambiguous, because the ENR associated with a node evolves over time. By pinning to a specific row in the 'record' table, we know the exact ENR at the time of the failure. It also makes the SELECT queries simpler, when looking up the sending client type.

The schema migration also generates the record data, by replacing the node ID with the most recent record ID at the time of migration. While technically inaccurate, it's better than losing the data, and probably not worth doing more detailed work to find the exact ENR at the time of the transfer for this relatively short window of legacy data.

---

I generated some fake data locally to test this against, including multiple records per node. I was able to migrate up and down repeatedly, without issues.

TODO:
- [x] grab the node record at the time of the transfer failure, to fill in the correct column 

Might as well review as one change instead of per-commit. This is relatively short, and all pretty interrelated.


carver:
Since we might not have collected a record yet for the given node (in the case that we stumble on the node during a transfer before we've found it during a census), it might be worth just keeping the node ID also, instead of removing the column. Does anyone have thoughts?

carver:
I'll give this one more pass, but I don't think there will be a ton of changes before it's ready for review.

carver:
Hm, the "low confidence" comment at least improves readability, and maybe fixes some scenario that I haven't come up with yet. So I'll add the constraint.

entity/src/audit_internal_failure.rs
```diff
@@ -14,6 +15,7 @@ pub struct Model {
     pub id: i32,
     pub audit: i32,
     pub sender_node: i32,
+    pub sender_record_id: Option<i32>,
```
mrferris, line 18: I think this should be non-optional because it's based off a QueryTrace object, whose ENR entry for the node is non-optional so we'll always have it, even in the test cases. Also prevents edge conditions eg ENRs changing between query time and log time.

So we can just pass the ENR object into the create function and look it up.

migration/src/m20250404_202958_internal_failures_replace_node_with_record.rs
```diff
@@ -0,0 +1,68 @@
...
+                    .to(Record::Table, Record::Id)
+                    .on_update(ForeignKeyAction::Cascade)
+                    .to_owned(),
+            )
+            .await?;
+
+        // For all internal transfer failures, infer the sender record from the from the sender node.
+        // Use the most recent record available for the node. This is helpful, even if wrong.
+        manager.get_connection().execute_unprepared(
+            "WITH maxsn AS ( SELECT r.*, ROW_NUMBER() OVER (PARTITION BY node_id ORDER BY sequence_number DESC) AS rn FROM record AS r) UPDATE audit_internal_failure AS aif SET sender_record_id=maxsn.id FROM maxsn WHERE aif.sender_node=maxsn.node_id AND maxsn.rn=1;",
```
mrferris, line 36: nice

mrferris (changes requested):

entity/src/audit_internal_failure.rs
carver: Ah whoops I forgot that the QueryTrace embedded the ENR. I'll go find/extract that.

The only other point of awkwardness is that at the time the column is added, the column is empty. So I guess the next best idea I have is to create the column as nullable, then migrate the data, then alter the column definition to make it non-null.

carver: Ok, got it. The `sender_node` is starting to look redundant. I guess I'll remove it next. I delayed a bit, because I didn't want to do the data migration to recover it on a `migrate down` call. But as I type it out, I don't think this is the good kind of lazy 😆 I guess it's worth it.

carver:
The PR is ready for your next look @mrferris

PR #389, opened 2025-04-16
Store and Display client info data from Ping Extensions

Ktl-XV:
Sanitizes data after saving it to the DB but before sending it to the web client, that way the DB has raw data for debugging while the web client only shows sanitized strings. 

A new areaGraph function was created to replace existing area graphs and allow for easier maintenance.

New Clients Page:
image


Ktl-XV:
Superseded by #390

Issue #382, opened 2025-03-24
Show protocol versioning statistics


pipermerriam:
### What is wrong

https://github.com/ethereum/portal-network-specs/pull/379

Protocol versioning is coming to portal and glados doesn't show any info about this.

### How can it be fixed.

We need a simple new table that shows the latest data about supported versions for each network.  The table should show:

- Number of nodes that do not have a `pv` entry in their ENR
- Number of nodes that support each protocol version.

Eventually we may want more nuanced information but this simple table should do for now.

behalshabnam:
@pipermerriam I would like to work on this, can you assign me this one?

pipermerriam:
We don't really assign issues.  If you'd like to work on this please just open a pull request.

PR #390, opened 2025-04-16
feat(web): add protocol version graph

Ktl-XV:
Add a graph to display the max supported protocol versions of nodes found in each census. 

~Nodes that did not published a protocol version in their ENR, are shown as version 0.~
Nodes that do not publish a protocol version in their ENR are not displayed in the graph, to better accommodate #382. Moved the "Nodes in Census" graph up so it is next to the new graph too

Lower part of the Network Overview page:
image


mrferris (approved):
⭐️

PR #396, opened 2025-04-30
feat: new internal transfer failure chart on dash

### LINKED ISSUE #336: Compute aggregate transfer success rate stats from the trace

carver:
Fix #336 

TODO:
- [x] test locally by generating some fake test data and confirming it show up in the chart

Gut check test with a couple different failure records locally:
image

This screenshot ^ is the combination of this PR, #403 and #404

### DIFF
--- a/glados-web/src/lib.rs
+++ b/glados-web/src/lib.rs
@@ -130,6 +130,10 @@ pub async fn run_glados_web(config: Arc<State>) -> Result<()> {
             "/census/census-node-timeseries-data/",
             get(routes::census_timeseries),
         )
+        .route(
+            "/api/weekly-transfer-failures/",
+            get(routes::weekly_transfer_failures),
+        )
         .route("/api/audit-block-status/", get(routes::audit_block_status))
         .nest_service("/static/", serve_dir.clone())
         .fallback_service(serve_dir)
--- a/glados-web/src/routes.rs
+++ b/glados-web/src/routes.rs
@@ -1562,7 +1562,7 @@ pub async fn weekly_transfer_failures(
         .all(&state.database_connection)
         .await
         .map_err(|e| {
-            error!(err=?e, "Failed to lookup census node timeseries by client protocol versions");
+            error!(err=?e, "Failed to lookup weekly transfer failures");
             StatusCode::INTERNAL_SERVER_ERROR
         })?;
     Ok(Json(transfer_failures))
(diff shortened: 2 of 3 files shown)


carver:
I've never seen this before:

```Downloading datasets/mainnet/blocks/premerge.csv (1.6 GB)
Error downloading object: datasets/mainnet/blocks/premerge.csv (fd63fd1): Smudge error: Error downloading datasets/mainnet/blocks/premerge.csv (fd63fd1b64233680eecbdc3903c951f9eea7ea43c366e5421fe62106039c5ed1): batch response: This repository is over its data quota. Account responsible for LFS bandwidth should purchase more data packs to restore access.

Errors logged to '/home/circleci/project/.git/lfs/logs/20250430T232212.101541832.log'.
Use `git lfs logs last` to view the log.
error: external filter 'git-lfs filter-process' failed
fatal: datasets/mainnet/blocks/premerge.csv: smudge filter lfs failed```

mrferris (approved):
LGTM, tested with prod data 👍

PR #397, opened 2025-05-01
chore: disable lfs checkout in lint-build-test CI

carver:
In #396 I noticed that CI was failing because the git checkout failed, because of going over the LFS quota. This isn't a complete solution, but at least unblocks testing on active PRs.


carver:
Hm, there was a `cargo fmt` check failure that makes no sense to me (it says a file is missing that is there from my last SSH into circle test). Hopefully something silly that I'll fix in the next couple of minutes.

mrferris (approved):

PR #402, opened 2025-05-08
fix: ignore lfs during CI docker builds

carver:
Also, make a reusable step for a cleaner circle config.

I'm not sure how to test this cleanly without a merge to master. I am not clear whether the built docker images will work without the premerge.csv. But I figure I'd get the PR started as a place to discuss.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -5,12 +5,32 @@ executors:
   docker-publisher:
     docker:
       - image: cimg/rust:1.85.0
+commands:
+  non_lfs_checkout:
+    steps:
+      - run:
+          name: Check out code, without lfs files
+          command: |
+            # add github ssh keys to known_hosts
+            mkdir -p ~/.ssh
+            curl -L https://api.github.com/meta | jq -r '.ssh_keys | .[]' | sed -e 's/^/github.com /' >> ~/.ssh/known_hosts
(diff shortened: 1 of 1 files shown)


KolbyML:
> I'm not sure how to test this cleanly without a merge to master.

image

Maybe merge queue's could be used? I use `Github Merge Queue` on other projects to make sure CI passes, before doing the merge itself. So it basically runs CI in a queue of PR's then merges them if it passes.

KolbyML (approved):
:shipit: from looking over this PR, I think CI would pass if this was merged. As the `non_lfs_checkout` command is working on `master` (before it was refactoed) and it is working for the `lint-build-test` CI ran on this PR.

Good work :rocket:

KolbyML:
I am a little surprised `premerge.csv` wasn't stored in a new/separate github repo tho. As it seems like the thing you would download when you need it, which is what Glados is currently doing, at least from what I understand.

carver:
> I am a little surprised `premerge.csv` wasn't stored in a new/separate github repo tho. As it seems like the thing you would download when you need it, which is what Glados is currently doing, at least from what I understand.

Yeah, I agree, it should probably be somewhere else, like S3, so that every repo checkout doesn't download it immediately by default.

PR #404, opened 2025-05-08
feat(web): areaGraph - show a default Other field

carver:
If data is sent by the server with a key not specified by seriesMetadata, then display it anyway with an Other name.


carver:
cc @Ktl-XV for review

Ktl-XV:
lgtm

PR #408, opened 2025-05-13
chore: enable beacon latest audits in docker compose

Ktl-XV:
Currently only the Random strategy runs for Beacon audits when using docker-compose.

This PR enables Latest Audits for Beacon


mrferris (approved):
LGTM

PR #426, opened 2025-05-20
feat: add Samba

KolbyML:
# Issue

Samba is listed as Unknown on Glados

# Solution

Add Samba.

Shisui was missing from a few places as well so I added them

PR #427, opened 2025-05-21
fix: strum props name used Shisui instead of Samba

KolbyML:
@carver pointed out I accidentally used the word "shisui" instead of samba in one place here is a PR to fix that. https://github.com/ethereum/glados/pull/426


carver (approved):
🚢

PR #435, opened 2025-05-23
perf(web): load radii of peers in audit details

### LINKED ISSUE #431: Audit detail page is slow
Looking up the node radius seems to be the culprit. For example, look at this 17 second load time...

```
[2025-05-22T21:00:56Z WARN  sqlx::query] SELECT DISTINCT ON (n.node_id) …; rows affected: 37, rows returned: 37, elapsed: 17.159s
    
    SELECT
      DISTINCT ON (n.node_id) n.node_id,
      cn.data_radius
    FROM
      node n
      JOIN record r ON r.node_id = n.id
      JOIN census_node cn ON cn.record_id = r.id
    WHERE
      n.node_id = ANY($1 :: bytea [])
    ORDER BY
      n.node_id,
      ABS(
        EXTRACT(
          EPOCH
          FROM
            (cn.surveyed_at - $2 :: timestamp)
        )
      )
```

carver:
Fix #431 

Load time went from 17 seconds to 170 ms.

Note that the behavior changes slightly: it pulls the most recent census data that is up to 15 minutes after the audit time.

This is faster because we can do the sorting & limiting of the census nodes all in index by using the `surveyed_at` value directly, instead of the computed absolute difference.

Note that although we LIMIT 1 inside the census_node lookup, that is only limiting by record ID. Because each node may have multiple records, we only want the latest record for each node. So we still have to order by the time and apply DISTINCT node_id.

### DIFF
--- a/glados-web/src/routes.rs
+++ b/glados-web/src/routes.rs
@@ -618,16 +618,22 @@ pub async fn contentaudit_detail(
                 "
                 SELECT DISTINCT ON (n.node_id)
                     n.node_id,
-                    cn.data_radius
+                    closest_cn.data_radius
                 FROM
                     node n
                     JOIN record r ON r.node_id = n.id
-                    JOIN census_node cn ON cn.record_id = r.id
+                    CROSS JOIN LATERAL (
+                        SELECT cn.data_radius, cn.surveyed_at
+                        FROM census_node cn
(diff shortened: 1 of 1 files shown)


mrferris (approved):
🔥

PR #436, opened 2025-06-14
Remove Non-Working URL

Daulox92:
I removed the non-working URL for Nimbus mainnet. If you have an updated URL, please provide it, and I'll be happy to replace it.


morph-dev:
I believe that URL is correct, but it requires credentials to access it.

Because of that and the fact that Glados is no longer actively maintained project, I'm closing this PR.

